Issue #15, opened 2020-07-12
use interrogate, darglint, and pydocstyle in CI


MarcoGorelli:
Docs are really important and should be treated as such

PR #91, opened 2020-07-23
CI DOC Interrogate

MarcoGorelli:
xref #15


MarcoGorelli:
gosh...is there no way to just specify where to run it against?

PR #127, opened 2020-07-29
BUG allow for tools' usual configs to be used

### LINKED ISSUE #126: BUG nbqa doesn't pick up existing config files
e.g. from `.flake8`, `.isort`, etc.

MarcoGorelli:
closes #126


nbqa/__main__.py
```diff
@@ -19,6 +19,8 @@
     save_source,
 )
 
+CONFIG_FILES = {"flake8": [".flake8"]}
```
MarcoGorelli, line 22: need to update this as extensively as possible

PR #177, opened 2020-08-30
BUILD ignore build of docs in package

MarcoGorelli:


MarcoGorelli:
need to think about this

PR #199, opened 2020-09-09
Replacing the type annotation "re.Match" with Match from the typing m…

girip11:
Replacing the type annotation "re.Match" with type alias `Match` from the `typing` module in python

Reference:
https://docs.python.org/3.6/library/typing.html#typing.Match

Changes to be committed:
	modified:   nbqa/__main__.py

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -11,7 +11,7 @@ import tempfile
 from pathlib import Path
 from shlex import split
 from textwrap import dedent
-from typing import Dict, Iterator, List, Optional, Set, Tuple
+from typing import Dict, Iterator, List, Match, Optional, Set, Tuple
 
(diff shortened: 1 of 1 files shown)


MarcoGorelli (approved):
Thanks @girip11 ! I didn't know this was the correct way to do it, so this is much appreciated!

girip11:
1. Any reason, pylint is not included as part of .pre-commit-config.yaml. 
2. I could see two hooks for mypy as part of the pre-commit configuration. Any reason for two mypy hooks?

I am planning to add pylint to pre-commit config and remove one of the mypy precommit hooks from the config. Your thoughts on these changes?

MarcoGorelli:
@all-contributors please add @girip11 for code

Issue #124, opened 2020-07-29
DOC, TST: support pylint


MarcoGorelli:
Works really well out-of-the-box

MarcoGorelli:
I will have to make sure the pylint config file works

girip11:
@MarcoGorelli  Any reason why we name the temporary python file  with three spaces before the `.py` extension?

girip11:
With pylint, I am hitting two issues arising due to nbqa's internal working.

1.  Module name "notebook_with_indented_magics   " doesn't conform to snake_case
2.  unnecessary pass statement

If we can remove the additional spaces introduced in the file name, first issue can be fixed.
Second issue can be fixed, If I use some builtin function like `type()` or `id()`instead of `pass`. `print` is not preferred because of the issue with doctest.

 I am open to ideas

MarcoGorelli:
> @MarcoGorelli Any reason why we name the temporary python file with three spaces before the `.py` extension?

This is so that its name is of the same length as the notebook's name, in case any output uses the length of the file to make some fancy formatting box. Could use `aaa` insead of `   `?

As for `pass` / `print` - you'd suggested conditionally using `print`, depending on the tool which is used. So, `print` for pylint, and `pass` for the rest?

girip11:
Yes I had proposed different statements for different tools. But I am hesitant now to implement that way. I would like nbqa to be tool agnostic as much as possible. If we can avoid such tool dependent options, then we should go for it. What do you think?

girip11:
I used `type` builtin function pylint, flake8. It worked. To be sure I can run it once against all other supported tools.

MarcoGorelli:
Awesome! Thanks for looking into this, if `type()` works well with all the tools we're currently testing then it seems like a winner!

girip11:
I have one other solution. I am trying out that. I will try to send a sample PR in a day or two.

girip11:
PR https://github.com/nbQA-dev/nbQA/pull/302 to complete pylint support.

girip11:
Pylint support added. Closing this issue.

PR #201, opened 2020-09-09
Update the checkout and setup-python github actions from v1 to v2.

girip11:
* **checkout@v2** has improved performance over **checkout@v1** github action
* Similarly its recommended to update from **setup-python@v1** to v2
* Made python3 as the version to be used by all the pre-commit hooks globally rather than explicitly mentioning the python version in each of the precommit hooks.

Changes to be committed:
	modified:   .github/workflows/pre-commit.yml
	modified:   .pre-commit-config.yaml

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -9,12 +9,12 @@ jobs:
   pre-commit:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - uses: actions/setup-python@v1
-    - name: set PY
-      run: echo "::set-env name=PY::$(python -c 'import hashlib, sys;print(hashlib.sha256(sys.version.encode()+sys.executable.encode()).hexdigest())')"
-    - uses: actions/cache@v1
-      with:
-        path: ~/.cache/pre-commit
-        key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
-    - uses: pre-commit/action@v1.1.0
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - name: set PY
+        run: echo "::set-env name=PY::$(python -c 'import hashlib, sys;print(hashlib.sha256(sys.version.encode()+sys.executable.encode()).hexdigest())')"
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+      - uses: pre-commit/action@v1.1.0
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -11,10 +11,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - name: set PY
-        run: echo "::set-env name=PY::$(python -c 'import hashlib, sys;print(hashlib.sha256(sys.version.encode()+sys.executable.encode()).hexdigest())')"
-      - uses: actions/cache@v2
-        with:
-          path: ~/.cache/pre-commit
-          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
-      - uses: pre-commit/action@v1.1.0
+      - uses: pre-commit/action@v2.0.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,41 +1,43 @@
+default_language_version:
+  # force all unspecified python hooks to run python3
+  python: python3
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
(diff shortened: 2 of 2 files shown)


MarcoGorelli:
Thanks for this, @girip11 !

I remember Anthony Sottile (who authored pre-commit) saying there was some issue with using github actions v2 with pre-commit, although the check seems to be working here, so I guess it's been fixed

.github/workflows/pre-commit.yml
```diff
@@ -9,12 +9,12 @@ jobs:
...
-    - uses: pre-commit/action@v1.1.0
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - name: set PY
+        run: echo "::set-env name=PY::$(python -c 'import hashlib, sys;print(hashlib.sha256(sys.version.encode()+sys.executable.encode()).hexdigest())')"
+      - uses: actions/cache@v2
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+      - uses: pre-commit/action@v1.1.0
```
MarcoGorelli, line 20: Looks like there's a new version of this action: https://github.com/pre-commit/action

Do you want to update to v2?

MarcoGorelli (changes requested):

.github/workflows/pre-commit.yml
girip11: Sure, I will try out precommit v2 and update if there arent any issues.

MarcoGorelli: Also, as you correctly pointed out, there's an extra mypy hook here, so that one can be removed

s-weigand:
I don't know about the pre-commit@v1 action, but pre-commit@v2 takes care of caching itself.

s-weigand:
If you want to keep the versions of the used actions up to date you could also use [dependabot](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem), it also supports updates for pypi packages and part of github. I use it in projects where changing runtime dependencies might introduce breaking changes.
Even so `flake8`, `black` ... are optional runtime dependencies, having a heads up on breaking chages as soon as a new version gets released might be usefull.

MarcoGorelli:
Yes, you're right, the pre-commit action v2 is much simpler: https://github.com/pre-commit/action

girip11:
I am thinking of combining pylint steps from **pylint.yml** into precommit.yml github workflow since it makes sense to run all of the linting together in the same workflow. Suggested changes to pre-commit.yml

```YAML
jobs:
  pre-commit:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-python@v2
      - uses: pre-commit/action@v2.0.0
    - name: install-pylint
      run: python -m pip install --upgrade pip pylint
    - name: run-pylint
      run: pylint nbqa tests
```

This could seem like a primitive question. But is there any reason why pylint is not part of the requirements-dev.txt?

s-weigand:
> I saw that precommit hook for pylint is deprecated.

Are you sure about this? [`pre-commit-hooks` deprecated pyflakes](https://github.com/pre-commit/pre-commit-hooks#deprecated--replaced-hooks), but [`pylint ` has its own pre-commit hook](https://github.com/PyCQA/pylint).

Performance wise and for quality assurance, I think it would be best to add `pylint` to ` .pre-commit-config.yaml`.
That way `pylint` gets run each commit and `pre-commit/action@v2.0.0` will take care of caching + the workflow gets shorter.

> This could seem like a primitive question. But is there any reason why pylint is not part of the requirements-dev.txt?

If we use `pylint` with `pre-commit`, `pre-commit` will take care of the dependencies for each hook and installs them each in a separate env, so there will be no need to also install it your dev env.
At least until #124 gets implemented.

girip11:
> Are you sure about this? [`pre-commit-hooks` deprecated pyflakes](https://github.com/pre-commit/pre-commit-hooks#deprecated--replaced-hooks), but [`pylint ` has its own pre-commit hook](https://github.com/PyCQA/pylint).

@s-weigand thanks for the clarification. I checked this repo [https://github.com/pre-commit/mirrors-pylint] and it is mentioned **"This mirror repository is deprecated, use pylint directly"**. I mistook its meaning(Use pylint directly outside of the pre-commit hooks). I checked the pylint repo, and it has the **.pre-commit-hooks.yml**. So we can use the pylint repo directly.

girip11:
@MarcoGorelli Thanks for clarifying

> If you also want to include pylint in the pre-commit config file so it can all get checked together (and remove it from its own separate check) then that would be fantastic!

I am thinking of closing this PR with these set of changes. Adding pylint to pre-commit change can be taken in a separate PR.
Also I could help out in closing the issue[#124](https://github.com/nbQA-dev/nbQA/issues/124)

MarcoGorelli:
Sounds great - let's merge this one as it is then, and if you want to take on #124, that's awesome!

MarcoGorelli (approved):

PR #210, opened 2020-09-10
Centralize configs

s-weigand:
This is a pure taste issue.
But IMHO it looks better when the repo root only contains as many files as needed.
On the other hand, it is easier to find the config you want, when it is in its own separate file.
This all is due to the mess of which tool supports which common config files (`setup.cfg`?, `tox.ini`?, `project.toml`? and custom file).

So @MarcoGorelli that decision in all up to you. 😉 

At least this PR will be good as a reference, in case you prefer the custom file way.


girip11:
I too like the idea of having all these config either in pyproject.toml or setup.cfg

1. Looks like [pylint 2.5.0 or above supports pyproject.toml and setup.cfg](https://github.com/carsongee/pytest-pylint/pull/107). So we could move .pylintrc to either pyproject.toml or setup.cfg

2. Also **.coveragerc** contents can also be moved to pyproject.toml. 

We could do above mentioned changes too if @MarcoGorelli  approves of this PR. @s-weigand your thoughts on the above mentioned changes.

s-weigand:
@girip11 Thanks couldn't find that `pylint` feature and I think the `.coveragerc` I did overlook 😅

s-weigand:
Hmmmm looks like that `project.toml` support is only for the `pytest` addon `pytest-pylint`, but not `pylint` itself.
And even that doesn't seem to support all options.

girip11:
I referred [pylint cmdline ref](http://pylint.pycqa.org/en/latest/user_guide/run.html#command-line-options) and [Documentation page for configuration file](https://github.com/PyCQA/pylint/issues/3181), I added this snippet to pyproject.toml

```toml
[tool.pylint.'MESSAGES CONTROL']
#   "C0330", # to work well with `black`
#   "R0801", # Unfortunately I do need some duplication, e.g. with imports
#   "W1510", # I need to run subprocess.run without check because it's OK if it fails
#   "E0401", # import error: I run pylint in a separate environment so it's OK
disable = "C0330, R0801, W1510, E0401"
```

I ran pylint and it seems to be working well. @s-weigand Could you confirm this from your side?

s-weigand:
Never mind `pylint` does support toml, but their docs ......
There are only [8 hits for `toml`](https://github.com/PyCQA/pylint/search?q=toml&unscoped_q=toml&type=code) in the code, including the docs
In the end [this issue](https://github.com/PyCQA/pylint/issues/3472) gave me the hint I needed 😅

s-weigand:
Comment racing condition 😄 
IMHO the [bracket syntax I used](https://github.com/nbQA-dev/nbQA/pull/210/commits/bb5184bef447673e07b985b19ae25446faaf6c24) looks more compelling.
It's a list and looks like a python list (+ the comments work this way).

girip11:
> Comment racing condition smile
> IMHO the [bracket syntax I used](https://github.com/nbQA-dev/nbQA/pull/210/commits/bb5184bef447673e07b985b19ae25446faaf6c24) looks more compelling.
> It's a list and looks like a python list (+ the comments work this way).

Hey I would have too done this way. I just wanted to quickly check if pylint supports toml. So I tried with C0330 and it worked, so I posted my reply. Please proceed with your changes.

girip11:
Its good to see all these tools converging to support pyproject.toml and setup.cfg.

s-weigand:
Yeah, one standard location to put all the configs and my comment from above gets invalid.
> On the other hand, it is easier to find the config you want, when it is in its own separate file.

Anyway, I think we got rid of all custom files and now it is all up to @MarcoGorelli 😉

MarcoGorelli:
Wow, this is amazing, thank you @s-weigand and @girip11 :+1: ! I too didn't like the "mess" of having all these separate config files, so I'm glad you've taken care of this.

You're welcome to merge each others PRs BTW, no need to wait for my approval

Issue #212, opened 2020-09-11
CLN can we not ignore E0401 for pylint?


MarcoGorelli:
Now that pylint isn't incompatible with the rest of the dependencies, is it OK to remove
```
    "E0401",  # import error: I run pylint in a separate environment so it's OK
```
from pyproject.toml ?

s-weigand:
Try removing it and run `pre-commit run -a`, if it passes it should be fine.

girip11:
If this rule is removed from the pylint conf, we would get import error. **pytest and other import dependencies** in the source/test files won't be available to the pylint module since this would be run in an isolated environment by the pre-commit.  

To fix this issue we have two options.

1. We need to make all our source and test dependencies available to pylint during pre-commit check using the **additional_dependencies** key **.pre-commit-config.yaml** for pylint would look like below. But this rule would keep failing as and when we take dependency on some new other 3rd party package.

```YAML
  - repo: https://github.com/pycqa/pylint
    rev: pylint-2.6.0
    hooks:
      - id: pylint
        exclude: "docs"
        additional_dependencies:
          - "pytest"
```

2. Alternatively, we can make pylint as a local hook as suggested by creator the pre-commit. This is a generic change. We can freely add new dependencies and no need to change the pre-commit config everytime.

> pre-commit runs pylint from an isolated virtualenv. Many of pylint's checks perform dynamic analysis which will fail there. You may find configuring pylint as a local hook more useful.

I tried making pylint as a local hook with **language: system**. This will make the pylint picked up from our virtual environment or some other local environment based on PATH variable. When we develop, usually we have the virtual environment activated, so pre-commit will have not issue finding and running pylint. In github actions, we need to install all the dependencies for pylint to work under system flag.

3. Don't remove the rule E0401 from pylint configuration.

I could not think of any other alternate solution. I will raising PR. If you folks don't like the PR, feel free to close the PR without merging. [PR](https://github.com/nbQA-dev/nbQA/pull/229)

girip11:
PR merged, so closing this issue

PR #229, opened 2020-09-12
Remove pylint config E0401 and make pylint a local pre-commit hook

girip11:
Addresses the issue [#212](https://github.com/nbQA-dev/nbQA/issues/212)

If this rule **E0401** is removed from the pylint conf, we would get import error. **pytest and other import dependencies** in the source/test files won't be available to the pylint module since this would be run in an isolated environment by the pre-commit.  

To fix this issue we have 2 options.

1. We need to make all our source and test dependencies available to pylint during pre-commit check using the **additional_dependencies** key **.pre-commit-config.yaml** for pylint would look like below. But this rule would keep failing as and when we take dependency on some new other 3rd party package.

```YAML
  - repo: https://github.com/pycqa/pylint
    rev: pylint-2.6.0
    hooks:
      - id: pylint
        exclude: "docs"
        additional_dependencies:
          - "pytest"
```

2. Alternatively, we can make pylint as a local hook as suggested by creator the pre-commit. This is a generic change. We can freely add new dependencies and no need to change the pre-commit config everytime.

> pre-commit runs pylint from an isolated virtualenv. Many of pylint's checks perform dynamic analysis which will fail there. You may find configuring pylint as a local hook more useful.

I tried making pylint as a local hook with **language: system**. This will make the pylint picked up from our virtual environment or some other local environment based on PATH variable. When we develop, usually we have the virtual environment activated, so pre-commit will have not issue finding and running pylint. In github actions, we need to install all the dependencies for pylint to work under system flag before the precommit step is run.

3. Don't enable pylint rule E0401.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -11,4 +11,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
+      # This is required since pylint is used in pre-commit as a local hook
+      - name: install-pylint
+        run: python -m pip install --upgrade pip pylint
       - uses: pre-commit/action@v2.0.0
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -12,6 +12,8 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
       # This is required since pylint is used in pre-commit as a local hook
-      - name: install-pylint
-        run: python -m pip install --upgrade pip pylint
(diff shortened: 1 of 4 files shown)


MarcoGorelli (approved):
This looks great, thanks Girish! Am just wondering if there's a way to cache the dependencies so they don't need to be downloaded each time, but that's probably a separate issue

girip11:
Yes now that I have introduced a step to install dependencies outside of precommit, caching needs to be done.

MarcoGorelli:
cool, I've opened a separate issue for that, #231 - thanks!

PR #236, opened 2020-09-13
CONFIG Use setup.cfg instead of .nbqa.ini

### LINKED ISSUE #211: ENH Add support for project.toml
See discussion in #210 

A possible example, [corresponding to the docs](https://nbqa.readthedocs.io/en/latest/configuration.html) would be:
```toml
[tool.nbqa.mutate]
black = true
isort = false

[tool.nbqa.addopts]
flake8 = [
    "--ignore W503",
    ]

[tool.nbqa.configs]
mypy = ".mypy.ini"
```

MarcoGorelli:
closes #211 

EDIT
----

Looking at the issue, I think Sebastian's suggestion is better, i.e. instead of
```
[nbqa.flake8]
config=.flake8

[nbqa.black]
config=pyproject.toml
```
to have
```
[nbqa.config]
flake8=.flake8
black=pyproject.toml
```


s-weigand:
I think the approach with `[nbqa.config]`, makes it more obvious that this is a `nbQA` only setting, so users will be less tempted to duplicate their tool configuration in there.
And it would also couple it more tightly with the CLI flag name.

Also, refactoring out some code from `_get_configs` so `config_file` and `config_prefix`, can be used as arguments would allow to easily extend it to `project.toml` support and we could keep the `.nbqa.ini` for now, throwing a `DeprecationWarning`, instead of breaking it.

girip11:
I have not looked at the changes yet. Since nbqa is going to support **setup.cfg** for configuration instead of its own **nbqa.ini**, I suggest removing the **addopts** from the configuration. Let the users configure the tool's configuration either in `pyproject.toml` or `setup.cfg` separately from nbqa's configuration.

MarcoGorelli:
I was looking into how pydocstyle do it, and they support many different config files, for which they give an order of preference. So, they check one for a [pydocstyle] section, and if they find it, they don't check the rest. https://github.com/PyCQA/pydocstyle/blob/master/src/pydocstyle/config.py

Seems quite simple, so perhaps it's OK to support all 3. I would suggest:

1. PR to "officially" support setup.cfg and tox.ini (adding this one is no extra effort if we do it like pydocstyle do it), though silently keep supporting .nbqa.ini
2. Another PR to add support for pyproject.toml with `toml` as an optional dependency (I'd imagine parsing toml is a bit different, so this may take a little more work?)

MarcoGorelli:
> I suggest removing the **addopts** from the configuration. Let the users configure the tool's configuration either in `pyproject.toml` or `setup.cfg` separately from nbqa's configuration.

True, depends if there are tools which have arguments which can only be passed from the cli

s-weigand:
IMHO I would keep the addops, since users might want to have a different config for Notebooks than normal *.py files.
I.e. it is very common to a variable or function call as last line in a cell for showing the value. But liners will tell you that the statement has no effect. While this is fine in Notebooks, in *.py files this would be a Codes smell.

girip11:
> True, depends if there are tools which have arguments which can only be passed from the cli
Agreed. But I am not sure if tools support CLI only argument.

> it is very common to a variable or function call as last line in a cell for showing the value.
Regarding this, using a statement like `print`, or `display` since **"Explicit is better than implicit"**
. Also when you add new line in the same cell, your intention of printing it has to be taken care by some other contributor or your future self. Though its a common behavior if its a smell, I feel it should be fixed.

@s-weigand I can understand your **intentions** here (maintaining a separate configuration for notebooks).  I see the python code in a jupyter notebook the same way as a python source code in terms of good practices. This makes it simple to start with.

In this case I would take the approach of the requirement coming from a user(in near future) rather than speculating the future if we don't have any compelling points now. This way we will have a broader view of the issue because we know how the users wanted the feature.

girip11:
Let's say we proceed with two places to define flake8's configuration. 
1. pyproject.toml under flake8 section
2.  setup.cfg under nbqa addopts.

Now we are getting in to the business of dealing with multiple configurations for the same tool. You might have to merge the configurations from both places and then feed it to flake8.

girip11:
Also if the developers need some exemption of a linting rule for a particular notebook, they should be encouraged to use a disable comment locally in the notebook (ex: `#noqa`) rather than nbqa tool trying the take care of these **disables**.

These are the reasons why I am up for removing this addopts argument. I am fine with however the way your proceed.

MarcoGorelli:
@girip11 thanks for sharing your thoughts very clearly :+1: ! From I've understood, your argument is: we should disable addopts because it allows users to have different configurations for notebooks than for python files, and that would be a bad practice. You also share concerns about it being complex to implement due to conflicts.

First, disabling addopts doesn't disallow having different configurations: you can still pass your own config file, e.g. by having this in setup.cfg:
```cfg
[nbqa.config]
flake8 = .my_flake8_config_file
```

Second, are you sure it's necessarily a bad practice? E.g. if running isort, it seems legitimate to want imports to not be sorted between cells for notebooks (e.g. with the `--treat-comment-as-code`flag) but not to want that for Python files.

Third, regarding conflicts: these are handled by the third-party tool itself. For example, if you run have setup.cfg like
```
[nbqa.config]
flake8 = .flake8
addopts = -q
```
, then nbqa will copy your `.flake8` file over to your temporary directory and then run
```
flake8 <python file> -q
```
. So what happens if you also have `-v` in `.flake8`, or something incompatible with it? Well, I don't know, but we can let flake8 deal with that, just like it normally would when it runs against Python files.

So, IMO it's OK to keep supporting it. If you disagree or if I've misrepresented your position, please do let me know!

girip11:
@MarcoGorelli: You have captured my points clearly. I am agreeing with both of you to continue to support the addopts.

nbqa/__main__.py
```diff
@@ -451,6 +455,73 @@ def _run_command(
...
+    """
+    config_files = CONFIG_FILES
+    for config_file in config_files:
+        if config_file == "pyproject.toml":
+            continue  # Will be supported in a future PR.
+        config = configparser.ConfigParser()
+        config.read(config_file)
+        for section in NBQA_CONFIG_SECTION:
+            if config.has_section(f"nbqa.{section}"):
+                return config_file, config, "nbqa."
```
girip11, line 486: Could we use some CONSTANT for storing the prefix"nbqa."

nbqa/__main__.py
```diff
@@ -451,6 +455,73 @@ def _run_command(
...
+    config = configparser.ConfigParser()
+    config.read(".nbqa.ini")
+    if config.has_section(command):
+        return ".nbqa.ini", config, ""
```
girip11, line 490: Could we refactor to use a CONSTANT like `DEFAULT_CONFIG_FILE = ".nbqa.ini"`

nbqa/__main__.py
```diff
@@ -451,6 +455,73 @@ def _run_command(
...
+    -------
+    config_file
+        Name of config file.
+    config
+        ConfigParser object which has read in config file.
+    config_prefix
+        Prefix of sections in config file.
+    """
+    config_files = CONFIG_FILES
+    for config_file in config_files:
```
girip11, line 479: We could use CONFIG_FILES directly in the for loop instead of `config_files = CONFIG_FILES`

nbqa/__main__.py
```diff
@@ -451,6 +455,73 @@ def _run_command(
...
+        e.g. config, mutate, ...
+    option
+        e.g. flake8, black, ...
+
+    Returns
+    -------
+    str
+        Parsed option.
+    """
+    if config_file != ".nbqa.ini":
```
girip11, line 516: Querying from different configurations `nbqa.ini`, `setup.cfg`, `pyproject.toml` seems to be a good place to apply polymorphism, Different configuration handlers exposing same interface to query the config. We can avoid these if/else structures. @MarcoGorelli Your thoughts on this? Do you have some sort of preference in using classes?

tests/conftest.py
```diff
@@ -10,6 +10,23 @@
...
+    """
+    Temporarily delete setup.cfg so it can be recreated during tests.
+
+    Parameters
+    ----------
+    tmpdir
+        Pytest fixture, gives us a temporary directory.
+    """
+    filename = Path("setup.cfg")
+    temp_file = Path(tmpdir) / "setup.cfg"
```
girip11, line 24: You could reuse `filename` in the definition of `temp_file`.

girip11 (changes requested):

MarcoGorelli:
Thanks @girip11  for your (as always) excellent comments, and for holding me accountable to not writing lazy code :+1: 

Regarding polymorphism - IMO at the moment it's not too complex that we would _need_ it, but if you wanted to submit a PR for that it would be welcome!

girip11:
> Thanks @girip11 for your (as always) excellent comments, and for holding me accountable to not writing lazy code +1
> 
> Regarding polymorphism - IMO at the moment it's not too complex that we would _need_ it, but if you wanted to submit a PR for that it would be welcome!

Sure once these PRs get merged, I would like to take a look at this config refactoring

girip11 (approved):

PR #237, opened 2020-09-13
By default, preserve setup.cfg, mypy.ini, and pyproject.toml files

### LINKED ISSUE #232: CONFIG copy over setup.cfg and pyproject.toml by default if no config is specified

MarcoGorelli:
closes #232 

Last PR for the day :rofl: Once this batch is done I think we can release 0.2.0

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -306,13 +306,13 @@ def _preserve_config_files(nbqa_config: Optional[str], tmpdirname: str) -> None:
     config_files = ["setup.cfg", "pyproject.toml"]
     if nbqa_config is not None:
         config_files = [nbqa_config]
-    for config_file_ in config_files:
-        target_location = Path(tmpdirname).joinpath(
-            Path(config_file_).resolve().relative_to(Path.cwd())
+    for config_file in config_files:
+        target_location = Path(tmpdirname) / Path(config_file).resolve().relative_to(
+            Path.cwd()
         )
         target_location.parent.mkdir(parents=True, exist_ok=True)
         shutil.copy(
-            str(config_file_),
+            str(config_file),
             str(target_location),
         )
 
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -307,16 +307,13 @@ def _preserve_config_files(nbqa_config: Optional[str], tmpdirname: str) -> None:
     if nbqa_config is not None:
         config_files = [nbqa_config]
     for config_file_ in config_files:
-        Path(tmpdirname).joinpath(
+        target_location = Path(tmpdirname).joinpath(
             Path(config_file_).resolve().relative_to(Path.cwd())
-        ).parent.mkdir(parents=True, exist_ok=True)
+        )
+        target_location.parent.mkdir(parents=True, exist_ok=True)
         shutil.copy(
             str(config_file_),
-            str(
-                Path(tmpdirname).joinpath(
-                    Path(config_file_).resolve().relative_to(Path.cwd())
-                )
-            ),
+            str(target_location),
         )
 
 
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -15,6 +15,8 @@ from typing import Dict, Iterator, List, Match, Optional, Set, Tuple
 
 from nbqa import __version__, replace_source, save_source
 
+CONFIG_FILES = ["setup.cfg", "tox.ini", "pyproject.toml"]
+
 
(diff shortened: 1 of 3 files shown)


nbqa/__main__.py
```diff
@@ -303,19 +303,18 @@ def _preserve_config_files(nbqa_config: Optional[str], tmpdirname: str) -> None:
...
-            )
-        ),
-    )
+    config_files = ["setup.cfg", "pyproject.toml"]
+    if nbqa_config is not None:
+        config_files = [nbqa_config]
+    for config_file_ in config_files:
+        target_location = Path(tmpdirname).joinpath(
+            Path(config_file_).resolve().relative_to(Path.cwd())
+        )
```
s-weigand, line 312: Can be simplified to:
```python
target_location = Path(tmpdirname) / Path(config_file_).resolve().relative_to(Path.cwd()
```

tests/test_configs_work.py
```diff
@@ -112,3 +112,38 @@ def test_configs_work_in_nbqa_ini(
...
+        handle.write(
+            dedent(
+                """\
+            [flake8]
+            ignore=F401
+            select=E303
+            quiet=1
+            """
+            )
+        )
```
s-weigand, line 136: Can be simplified to:
```python
Path("setup.cfg").write_text("[flake8]\nignore=F401\nselect=E303\nquiet=1")
```

s-weigand (changes requested):

nbqa/__main__.py
girip11: Could you rename `config_file` to `config_file` so that it follows the snake_case convention

tests/conftest.py
```diff
@@ -10,6 +10,23 @@
...
+    """
+    Temporarily delete setup.cfg so it can be recreated during tests.
+
+    Parameters
+    ----------
+    tmpdir
+        Pytest fixture, gives us a temporary directory.
+    """
+    filename = Path("setup.cfg")
+    temp_file = Path(tmpdir) / "setup.cfg"
```
girip11, line 24: `temp_file = Path(tmpdir) / filename` so that setup.cfg is hard coded at only one place.

tests/test_configs_work.py
girip11: @s-weigand I find the `dedent`version more readable compared to the your simplified code snippet. But I agree to use `Path.write_text`. So I suggest, `Path.write_text(dedent(...))`

s-weigand: I use [MagicPython](https://github.com/MagicStack/MagicPython) for syntax highlighting, which also works on strings, so for me it is quite readable. But yeah, if is more than 2 lines, than the decent might be more readable.

nbqa/__main__.py
```diff
@@ -303,19 +305,19 @@ def _preserve_config_files(nbqa_config: Optional[str], tmpdirname: str) -> None:
...
-    shutil.copy(
-        str(nbqa_config),
-        str(
-            Path(tmpdirname).joinpath(
-                Path(nbqa_config).resolve().relative_to(Path.cwd())
-            )
-        ),
-    )
+    if nbqa_config is not None:
+        config_files = [nbqa_config]
```
girip11, line 309: Here we expect the config name to be the default ones looked up by the tool right? For instance, if we are using flake8, then we set `config = .flake8`. But if we name it to something like `.flake8_nb` and give it to nbqa, it won't work, even though we copy the file to the tmp directory. I don't see the config parameter being passed as a command line parameter to the CLI command. flake8 has a command line parameter `--config` to specify a custom configuration file.

```Python
    output = subprocess.run(
        ["python", "-m", command, str(arg), *kwargs],
        stderr=subprocess.PIPE,
        stdout=subprocess.PIPE,
        cwd=tmpdirname,
        env=env,
    )
```

I know we can use addopts to specify `--config .flake8_nb`. But from a user's perspective the user would expect it to work by just setting the `--nbqa-config` setting right?Should it be considered a bug?

girip11 (changes requested):

nbqa/__main__.py
MarcoGorelli: @girip11 that's a really good point, thanks...I think we should just document that this needs to be one of the tool's standard configuration files and that users should consult the tools' docs to find out how to name/use them

girip11 (approved):

MarcoGorelli:
Massive thanks for your reviews! Have merged master, merging now following Girish's approval

PR #239, opened 2020-09-14
DOC fix config file name

MarcoGorelli:


MarcoGorelli:
Such a small fix I'll merge directly :) Feel free to do so yourselves as well, I'd just ask that we always open PRs though to make sure we don't accidentally make changes we didn't intend to

PR #247, opened 2020-09-20
BUG nbqa black removes trailing semi-colon

### LINKED ISSUE #241: BUG? nbqa black removes trailing semi-colon
e.g. after a plot

MarcoGorelli:
closes #241

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -443,6 +443,13 @@ def _run_command(
 
     before = _get_mtimes(arg)
 
+    if command == "black":
+        trailing_commas = [
+            n
+            for n, i in enumerate(arg.read_text().splitlines())
+            if i.strip().endswith(";")
+        ]
+
     output = subprocess.run(
         ["python", "-m", command, str(arg), *cmd_args],
         stderr=subprocess.PIPE,
@@ -453,6 +460,16 @@ def _run_command(
 
     after = _get_mtimes(arg)
     mutated = after != before
+
+    if command == "black" and mutated:
+        arg.write_text(
+            "\n".join(
+                [
+                    i if n not in trailing_commas else i + ";"
+                    for n, i in enumerate(arg.read_text().splitlines())
+                ]
+            )
+        )
 
     output_code = output.returncode
 
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -2,7 +2,6 @@
 
 import argparse
 import configparser
-import fileinput
 import os
 import re
 import shutil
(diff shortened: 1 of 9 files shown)


nbqa/__main__.py
```diff
@@ -451,9 +460,14 @@ def _run_command(
...
     )
 
-    after = _get_mtimes(arg)
-    mutated = after != before
+    mutated = _get_mtimes(arg) != before
 
+    if command == "black" and mutated and trailing_semicolons:
+        for line_number, line in enumerate(fileinput.FileInput(str(arg), inplace=True)):
+            if line_number in trailing_semicolons:
+                print(f"{line.rstrip()};{os.linesep}", end="")
```
girip11, line 468: Suppose I have a line whose length longer than the `max-line-length` configured for black. In such cases, black breaks a line in to multiple lines.

Before
```Python
print(" ".join(["When", "this", "line", "is", "longer", "than", "the", "black's", "line", "length"]));
```

After formatting

```Python
print(
    " ".join(
        [
            "When",
            "this",
            "line",
            "is",
            "longer",
            "than",
            "the",
            "black's",
            "line",
            "length",
        ]
    )
)
```

This logic would break in such cases. Please confirm with the tests

girip11 (changes requested):

nbqa/__main__.py
MarcoGorelli: @girip11 yes, you're absolutely right...I'm tempted to say that the solution, then is to only add a semicolon back in if it's at the end of a _cell_, rather than if it's at the end of a line

girip11: Shall we replace `;` with some magic string like `#SEMI` before formatting(preprocessing) and uncomment/replace #SEMI with `;` after formatting is done(post processing) ?

MarcoGorelli: I think we can keep track of which cells originally had trailing semi-colons and then add them back in if necessary, will update the PR now

nbqa/replace_source.py
```diff
@@ -32,28 +34,37 @@ def main(python_file: "Path", notebook: "Path") -> None:
...
+        cell_number
+            Number of current cell.
 
         Returns
         -------
         List[str]
             New source that can be saved into Jupyter Notebook.
         """
+        if cell_number in trailing_semicolons:
+            source = source.rstrip("\n") + ";"
```
MarcoGorelli, line 58: actually, this risks adding _two_ semicolons if the command doesn't remove semicolons...I should first check there isn't already a trailing comma

MarcoGorelli: wrote failing test, fixed now

nbqa/__main__.py
MarcoGorelli: @girip11 Have updated...thank you _so_ much for spotting the issue in my previous attempt at this, that would've been a bad bug :flushed:

girip11 (approved):

MarcoGorelli:
Just ran this on a load of notebooks from PyMC3 and it seems to be OK, so I'll merge - thanks for your help here!

Issue #95, opened 2020-07-24
CLN split __main__.py into smaller modules


MarcoGorelli:

girip11:
As part of this effort, I have raised this PR https://github.com/girip11/nbQA/blob/simple_config_refactoring/nbqa/config.py to separate configuration code out of `__main__.py`

s-weigand:
IMHO the `__main__.py` should only import a function and run it in `if __name__ == "__main__":`.
For me, the only legitimate reason to have definitions in `__main__.py`, is if it is the only module, besides `__init__.py` (which is the way `nbqa` started out at its birth, I guess).

As a side note, `__main__` also isn't picked up as a module, by `autodoc`.

MarcoGorelli:
this is arguably done by many of Girish's excellent refactoring PRs

PR #254, opened 2020-09-22
Nbqa configuration refactoring.

girip11:
This PR is ready for review but I would like to add more unit tests to config and config_parser modules

### DIFF
--- a/nbqa/config.py
+++ b/nbqa/config.py
@@ -1,33 +1,21 @@
 """Module responsible for storing and handling nbqa configuration."""
 
 from shlex import split
-from typing import Any, Callable, ClassVar, Dict, List, Optional
+from typing import Any, Callable, ClassVar, Dict, List, NamedTuple, Optional
 
 from nbqa.cmdline import CLIArgs
 
 
-class ConfigSections:  # pylint: disable=too-few-public-methods
+class _ConfigSections(NamedTuple):
     """Stores all the config section names."""
 
-    ADDOPTS: ClassVar[str] = "addopts"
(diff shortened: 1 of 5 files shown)


girip11:
This PR tries to address the issue https://github.com/nbQA-dev/nbQA/issues/95

MarcoGorelli:
Wow, this is awesome! I'm learning so much just by reviewing your work

> I would like to add more unit tests to config and config_parser modules

OK, so I should wait for them before merging?

nbqa/config.py
```diff
@@ -0,0 +1,129 @@
+"""Module responsible for storing and handling nbqa configuration."""
+
+from shlex import split
+from typing import Any, Callable, ClassVar, Dict, List, Optional
+
+from nbqa.cmdline import CLIArgs
+
+
+class ConfigSections:  # pylint: disable=too-few-public-methods
```
MarcoGorelli, line 9: Is it possible for the `sections` method to not belong to the class and just be a module-level function, so that this can just be a named tuple and we don't have to disable pylint warnings?

girip11:
> Wow, this is awesome! I'm learning so much just by reviewing your work

I am also learning a lot contributing to your work! 

> OK, so I should wait for them before merging?

Please proceed to merge these changes, since coverage shows 100%. I will send the tests for these modules as a separate PR.

nbqa/config.py
girip11: this is better.

MarcoGorelli:
Awesome! This seems correct to me...so, priority will be given to CLI args

PR #255, opened 2020-09-23
Support for pyproject.toml

girip11:
This is an optional PR. This tries to address https://github.com/nbQA-dev/nbQA/issues/211. **You are free to disapprove/reject this PR completely.** 

This makes toml as optional dependency. If toml is not present, then it won't parse from the pyproject.toml.

Since you mentioned in the issue

>That's a good suggestion, my only worry is that it would increase complexity, while having just a single file in which to look for configs would be easier to maintain...will think about it

I refactored the config part, I found it easier to add support to pyproject.toml with minimal code changes. So I went ahead and added those changes. I you want to support only `setup.cfg`, it's completely fine with me.


girip11:
As @s-weigand  suggested, I looked at how coverage implements toml as optional dependency and used the idea here.

docs/configuration.rst
```diff
@@ -63,9 +87,16 @@ To ignore extra cells, you can use the :code:`--nbqa-ignore-cells` CLI argument,
...
 
     [nbqa.ignore_cells]
     black = %%html,%%cython
+
+, or the :code:`ignore_cells` option in your :code:`pyproject.toml` file, e.g.
+
+.. code-block:: toml
+
+    [tool.nbqa.ignore_cells]
+    black = "%%html,%%cython"
```
MarcoGorelli, line 102: Wow, awesome!!

Does `toml` require the `%` to be escaped like `cfg` does? (I've opened #256 for the latter, which I forgot about when writing it :flushed: )

girip11: Thanks for pointing it out. I have no idea on this. Let me read about it and will address this comment accordingly.

girip11: For TOML, we don't need to escape `%`. I have added a test case to confirm the same.

MarcoGorelli:
> For TOML, we don't need to escape %. I have added a test case to confirm the same.

Wow, that's awesome! Thanks for looking into this. OK, I think that `pyproject.toml` should then be the default configuration file, as having to escape `%` is confusing. We can add `toml` as a required dependency, as it's a really small package anyway.

So, my suggestion would be:
- remove references to `setup.cfg` in the docs, and just have `pyproject.toml` as an example. I think that's easier than saying "otherwise, you can use setup.cfg, but you'll have to escape the percent signs"
- change `setup.py` so that `toml` is a required dependency

girip11:
Sure. I will make those changes and update the PR. 

>  OK, I think that pyproject.toml should then be the default configuration file, as having to escape % is confusing.

Support for `setup.cfg` and `tox.ini` would still be present in the code in addition to `pyproject.toml` right? Just that we would update the documentation to use pyproject.toml. Is my understanding correct?

MarcoGorelli:
Yes, if you agree I think we can still silently support them, but only list `pyproject.toml` in the docs

girip11:
Sure. We will support those other files. I will make the required changes.

docs/configuration.rst
```diff
@@ -12,28 +12,31 @@ If you wish to pass extra flags (e.g. :code:`--ignore W503` to :code:`flake8`) y
...
+    ]
 
-    [nbqa.addopts]
-    flake8 = --ignore W503
 
 Config file
 ~~~~~~~~~~~
 
-By default, :code:`nbQA` will use :code:`setup.cfg` to get the configs for your third-party tool.
+By default, :code:`nbQA` will look up the configuration of the third party tool in the following files :code:`setup.cfg`, :code:`tox.ini` and :code:`pyproject.toml`.
```
MarcoGorelli, line 28: I think we can just mention `pyproject.toml` here, if it's now a required dependency

nbqa/toml_parser.py
```diff
@@ -0,0 +1,44 @@
...
+def parse_from_pyproject_toml(command: str, file_path: Path) -> Optional[Configs]:
+    """
+    Parse nbqa configuration from TOML file.
+
+    Parameters
+    ----------
+    command : str
+        Third party tool to run
+    config_sections : List[str]
+        Configuration sections supported by nbqa e.g addopts, config, mutate
```
MarcoGorelli, line 22: missing parameter?

tox.ini
```diff
@@ -2,7 +2,10 @@
 envlist = py36, py37, py38
 
 [testenv]
-deps=-r requirements-dev.txt
+extras =
+    toml
```
MarcoGorelli, line 6: Is this still necessary?

MarcoGorelli (changes requested):
Awesome, have just left some minor comments

docs/configuration.rst
MarcoGorelli: nevermind me, this comment is correct, sorry :innocent:

MarcoGorelli (approved):
Nice, thanks a ton!

PR #276, opened 2020-09-27
Skip processing cells containing cell magics.

girip11:
Cell magics like %%script, %%bash and other custom cell magics configured via the ignore_cells configuration option will not be added to the temporary python file.

Changes to be committed:
	modified:   nbqa/__main__.py
	new file:   nbqa/notebook_info.py
	modified:   nbqa/replace_source.py
	modified:   nbqa/save_source.py
	modified:   setup.cfg

### DIFF
--- a/nbqa/replace_source.py
+++ b/nbqa/replace_source.py
@@ -5,20 +5,50 @@ The converted file will have had the third-party tool run against it by now.
 """
 
 import json
-from typing import TYPE_CHECKING, Dict, List
+from typing import TYPE_CHECKING, Iterator, List, Mapping, Set
 
+from nbqa.notebook_info import NotebookInfo
 from nbqa.save_source import CODE_SEPARATOR
 
 if TYPE_CHECKING:
     from pathlib import Path
 
 
-def main(
-    python_file: "Path",
-    notebook: "Path",
-    trailing_semicolons: List[int],
-    temporary_lines: Dict[str, str],
-) -> None:
+def _reinstate_magics(
+    source: str,
+    cell_number: int,
+    trailing_semicolons: Set[int],
+    temporary_lines: Mapping[str, str],
+) -> List[str]:
+    """
+    Put (preprocessed) line magics back in.
+
+    Parameters
+    ----------
+    source
+        Portion of Python file between cell separators.
+    cell_number
+        Number of current cell.
+    trailing_semicolons
+        List of cells which originally had trailing semicolons.
+    temporary_lines
+        Mapping from temporary lines to original lines.
+
+    Returns
+    -------
+    List[str]
+        New source that can be saved into Jupyter Notebook.
+    """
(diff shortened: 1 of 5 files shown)


girip11:
Refactored to skip writing the entire cell containing cell magics (default as well as the configured ones) to the temporary python file. After this change only the line magics will be parsed, replaced and written to the python file.

nbqa/replace_source.py
```diff
@@ -5,20 +5,50 @@
...
+    -------
+    List[str]
+        New source that can be saved into Jupyter Notebook.
+    """
+    rstripped_source = source.rstrip()
+    if cell_number in trailing_semicolons and not rstripped_source.endswith(";"):
+        source = rstripped_source + ";"
+    # we take [1:] because the first cell is just '\n'
+    for key, val in temporary_lines.items():
+        source = source.replace(key, val)
```
MarcoGorelli, line 47: Oops, looks like I put the comment  on the wrong line in a previous PR

```suggestion
    for key, val in temporary_lines.items():
        source = source.replace(key, val)
    # we take [1:] because the first cell is just '\n'
```

nbqa/save_source.py
```diff
@@ -9,51 +9,40 @@
...
-    ignore = MAGIC + [i.strip() for i in ignore_cells]
-
-    ignore_cell = source and any(source[0].lstrip().startswith(i) for i in ignore)
     for line in source:
         if (
-            line.lstrip().startswith("!") or line.lstrip().startswith("%")
-        ) or ignore_cell:
+            line.lstrip().startswith("!")
+            or line.lstrip().startswith("%")
+            or line.rstrip().endswith("?")
```
MarcoGorelli, line 40: Nice, hadn't thought of that!

MarcoGorelli (changes requested):
Legend, this is awesome! So much cleaner...just need to rush off now but will revisit later

nbqa/save_source.py
s-weigand: I think this can be refactored to:
```python 
if line.lstrip().startswith(("!", "%")) or line.rstrip().endswith("?"):
```
And since `?some_object` is also valid magic, it should be:
```python 
if line.lstrip().startswith(("!", "%", "?")) or line.rstrip().endswith("?"):
```

MarcoGorelli: > And since ?some_object is also valid magic,

there's no end to my surprises about Jupyter Notebooks

MarcoGorelli:
Awesome - timing results on pymc3s' docs:

on master:
```
$ time nbqa isort docs

real    0m0.988s
user    0m0.680s
sys     0m0.158s
```


on this branch
```
$ time nbqa isort docs

real    0m0.649s
user    0m0.594s
sys     0m0.057s
```

awesome work!!!

PR #277, opened 2020-09-27
DOC document pre-commit usage separately

MarcoGorelli:
Also, am changing to use `nbqa` (lowercase) when describing the package (unless it's in a title), as that's what you run from the command-line


MarcoGorelli:
I'll take the thumbs-up as approval :) Thanks!

PR #278, opened 2020-09-28
Handle indented magics

girip11:
Ready for review. This PR tries to handle the indented magics using `ast.parse`.

## Before
image

After running tools like **black** with --nbqa-mutate option, the source will have the line magics indented properly.

## After
image

### DIFF
--- a/nbqa/save_source.py
+++ b/nbqa/save_source.py
@@ -21,6 +21,19 @@ IGNORE_LINE_REPLACEMENT = "pass  # nbqa"
 
 
 def _is_src_code_indentation_valid(source: str) -> bool:
+    """
+    Return True is the indentation of the input source code is valid.
+
+    Parameters
+    ----------
+    source : str
+        Source code present in the notebook cell
+
+    Returns
+    -------
+    bool
+        True if the source indentation is valid otherwise False
+    """
     valid: bool = True
     try:
         ast.parse(source)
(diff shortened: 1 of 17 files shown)


MarcoGorelli:
Awesome!! I don't have time for a proper review now but I'll come back to this tomorrow - in the meantime, thank you so much!

nbqa/save_source.py
```diff
@@ -19,6 +20,73 @@
...
+    # preserve the leading spaces and check if the syntax is valid
+    line_tokenised = f"{leading_space}{IGNORE_LINE_REPLACEMENT}{token}"
+
+    if leading_space and not _is_src_code_indentation_valid(
+        "".join(source + [line_tokenised])
+    ):
+        # Remove the line magic indentation assuming these leading spaces
+        # lead the source to have invalid indentation
+        # Currently we don't check if the original source
+        # code itself had IndentationError
```
MarcoGorelli, line 80: That's a good point...if we're going to do this, reckon there would be much overhead in checking that the source code _without_ this extra line has no indentation error? Perhaps we need an extra test-case in which the indentation error doesn't come from the magic?

MarcoGorelli: On a second thought, what you're doing is probably alright - if the original source code itself has an indentation error, then any tool which uses ast.parse (e.g. `black`) will report an error anyway

MarcoGorelli (approved):
Looks good! I'll try running this on some pymc3 notebooks to see if/how performance is affected and unless there's some huge loss, I'll merge

Issue #279, opened 2020-09-29
Don't ignore in-line magic (when possible)


MarcoGorelli:
There are some in-line magics that _could_ be replaced with valid Python. For example
```
from foo import func

if True:
    %time func()
```

will currently get transformed to 
```
from foo import func

if True:
    pass # <comment with hash>
```
meaning that `flake8` would incorrectly warn that `func` is imported but unused.

Not sure how the substitution would work, but I think this would make sense as a longer-term goal

girip11:
The approach I used to solve this issue is to first extract only the python code from the statement. Then parse the python code snippet using ast.parse and check if the python code contains any Call node. I used Call node type since most of the time we either import a function or class. If the python code contains a Call node, I would place the entire code inside the template string "callable(lambda _: {code}) # {token}". Impl can be found here https://github.com/girip11/nbQA/blob/b83aaadb9b6f334a719d00ddc74abd18b396cb6d/nbqa/handle_magics.py#L15

I know this is not the best of approaches. Any suggestions are welcome.

girip11:
But with this approach, we would have issue replacing the statement back to the original statement when tools like pyupgrade, black modify the code.
For example

```Python
from foo import func

if True:
    %time func(arg = 'arg1')
```

After replacing with template becomes

```Python
from foo import func

if True:
    callable(lambda _: func(arg = 'arg1')  # token
```

When black runs on this code, we would get

```Python
from foo import func

if True:
    callable(lambda _: func(arg="arg1")  # token
```

This would affect the `replace_magics` logic, where we just where we currently check exact match assuming the replaced python code does not get altered.

I thought of adding comments to disable black (`#fmt: off`), pyupgrade. But I find this solution very ugly and won't scale on any new tools since nbqa is tool agnostic.

So we would have to find another way to replace the original magic back using some logic rather than just a string equality check.

girip11:
This is addressed by [PR](https://github.com/nbQA-dev/nbQA/pull/377). Closing this issue.

PR #284, opened 2020-10-02
Different approach to replace line magics with a valid python source …

girip11:
This is just a different approach to replace line magics with a valid python source code. Ignore the CI failures (failing due to code coverage). **This will not be commited**. This PR is just to demonstrate the approach.

Here we define a `no_op` function in nbqa itself, and calling that function in place of line magics using `__import__`. So the replacement python code will be `IGNORE_LINE_REPLACEMENT = '__import__("nbqa.util", fromlist=["nbqa"]).no_op("{}")'
` instead of `IGNORE_LINE_REPLACEMENT = "pass  # nbqa"`

* I don't think anyone will use this code in a notebook, since this code is internal to the `nbqa`.
* Actually the statement does nothing so no issues caused to tools like `doctest`.
* Passes all linter checks, mypy checks and formatters

Any comments/thoughts on this approach? Would it be an issue when used with some other tool?

### DIFF
--- a/nbqa/save_source.py
+++ b/nbqa/save_source.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
 
 CODE_SEPARATOR = "# %%"
 MAGIC = ["%%script", "%%bash"]
-IGNORE_LINE_REPLACEMENT = "pass  # nbqa"
+IGNORE_LINE_REPLACEMENT = '__import__("nbqa.util", fromlist=["nbqa"]).no_op("{}")'
 
 
 def _is_src_code_indentation_valid(source: str) -> bool:
@@ -69,7 +69,7 @@ def _handle_line_magic_indentation(
     token = secrets.token_hex(3)
 
     # preserve the leading spaces and check if the syntax is valid
-    line_tokenised = f"{leading_space}{IGNORE_LINE_REPLACEMENT}{token}"
+    line_tokenised = f"{leading_space}{IGNORE_LINE_REPLACEMENT.format(token)}"
 
     if leading_space and not _is_src_code_indentation_valid(
         "".join(source + [line_tokenised])
--- a/nbqa/util.py
+++ b/nbqa/util.py
@@ -1,0 +1,6 @@
+"""Module with helper utils to handle line magics in the notebook source."""
+
+
+def no_op(_: str) -> None:
+    """Do nothing."""
+    return


MarcoGorelli:
Nice idea! The only drawback I can think of is that, as there's an open bracket, then `black` might format it to be on multiple lines, e.g.
```
__import__(
    "nbqa.util", fromlist=["nbqa"]).no_op("{}"
)
```
would that be an issue?

girip11:
> Nice idea! The only drawback I can think of is that, as there's an open bracket, then `black` might format it to be on multiple lines, e.g.
> 
> ```
> __import__(
>     "nbqa.util", fromlist=["nbqa"]).no_op("{}"
> )
> ```
> 
> would that be an issue?

Black would format/split only when the line length exceeds the configured max line length right?

MarcoGorelli:
> Black would format/split only when the line length exceeds the configured max line length right?

Yes, exactly, but that's my concern, we don't know what `--line-length` it's being called with. If the magic has a single indent, then we should be fine, but if the magic happens in a block which is indented several times them it may be a problem?

Is there any drawback to using `type()`?

girip11:
Black by default has line length at 88 characters. `__import__("nbqa.util", fromlist=["nbqa"]).no_op("e2d4d4")` takes 61 characters. Even with 5 levels of indentation (4 spaces per indentation), line length is well under the limit.

I am thinking, with this approach, we should be able to address https://github.com/nbQA-dev/nbQA/issues/279(with some code changes)

```Python
from foo import func

if True:
    %time func()
```

can be transformed to

```Python
from foo import func

if True:
    # Even if black formats this in to multiple lines, we a line delimiting comment 
    # like nbqa, we can handle such scenarios.
    __import__("nbqa.util", fromlist=["nbqa"]).no_op("e2d4d4", func()) # nbqa
```

The point is, with possibility of writing new methods to handle such issues

girip11:
> Is there any drawback to using type()?

I can't think of any drawback using `type()`. Just thought of sharing this approach and see if it's any better than using `type()`. I am fine with either approach.

MarcoGorelli:
That's a really good point! If this has the possibility of being extended to address #279 , then I'm happy if our package uses this new approach directly rather than first transitioning to `type()` and then transitioning to something else

I'll edit setup.py to list you as co-author - this library started as a little hack of mine, but with recent developments I think it's really becoming something to be proud of and this wouldn't have been possible without you

girip11:
> That's a really good point! If this has the possibility of being extended to address #279 , then I'm happy if our package uses this new approach directly rather than first transitioning to type() and then transitioning to something else

Then I will refactor accordingly and start send out the PRs


> I'll edit setup.py to list you as co-author - this library started as a little hack of mine, but with recent developments I think it's really becoming something to be proud of and this wouldn't have been possible without you.

Thanks a lot @MarcoGorelli. I really liked this project idea of yours, I am happy that I could contribute.

girip11:
closing this PR

PR #286, opened 2020-10-02
Simplify output parsing

MarcoGorelli:
Somehow I made this _way_ too complicated the first time around :flushed: :rofl:

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -69,6 +69,44 @@ def _temp_python_file_for_notebook(
     return temp_python_file
 
 
+def _replace_full_path_out_err(
+    out: str, err: str, temp_python_file: Path, notebook: Path
+) -> Tuple[str, str]:
+    """
+    Replace references to temporary Python file's full path with notebook's path.
+
+    Parameters
+    ----------
+    out
+        Captured stdout from third-party tool.
+    err
+        Captured stderr from third-party tool.
+    temp_python_file
+        Temporary Python file where notebook was converted to.
+    notebook
+        Original Jupyter notebook.
+
+    Returns
+    -------
+    out
+        Stdout with temporary Python file's full path with notebook's path.
+    err
+        Stderr with temporary Python file's full path with notebook's path.
+    """
+    out = out.replace(str(temp_python_file), str(notebook))
+    err = err.replace(str(temp_python_file), str(notebook))
+
+    out = out.replace("   .py", ".ipynb")
+    err = err.replace("   .py", ".ipynb")
+
+    # This next part is necessary to handle cases when `resolve` changes the path.
+    # I couldn't reproduce this locally, but during CI, on the Windows job, I found
+    # that VSSADM~1 was changing into VssAdministrator.
+    out = out.replace(str(temp_python_file.resolve()), str(notebook.resolve()))
+    err = err.replace(str(temp_python_file.resolve()), str(notebook.resolve()))
+    return out, err
+
+
 def _map_python_line_to_nb_lines(
     out: str, notebook: Path, cell_mapping: Mapping[int, str]
 ) -> str:
@@ -108,7 +146,7 @@ def _map_python_line_to_nb_lines(
 
 
 def _replace_temp_python_file_references_in_out_err(
-    tempdirname: str,
+    temp_python_file: Path,
     notebook: Path,
     out: str,
     err: str,
@@ -119,8 +157,8 @@ def _replace_temp_python_file_references_in_out_err(
 
     Parameters
     ----------
-    tempdirname
-        Name of temporary directory.
+    temp_python_file
+        Temporary Python file where notebook was converted to.
     notebook
         Original Jupyter notebook.
     out
@@ -137,16 +175,7 @@ def _replace_temp_python_file_references_in_out_err(
     err
         Stderr with temporary directory replaced by current working directory.
     """
-    out = (
-        out.replace(f"{tempdirname}{os.sep}", "")
-        .replace(f"{str(Path(tempdirname).resolve())}{os.sep}", "")
-        .replace("   .py", ".ipynb")
-    )
-    err = (
-        err.replace(f"{tempdirname}{os.sep}", "")
-        .replace(f"{str(Path(tempdirname).resolve())}{os.sep}", "")
-        .replace("   .py", ".ipynb")
-    )
+    out, err = _replace_full_path_out_err(out, err, temp_python_file, notebook)
     out = _map_python_line_to_nb_lines(out, notebook, cell_mapping)
     return out, err
 
@@ -417,7 +446,7 @@ def _run_on_one_root_dir(
 
         for notebook, temp_python_file in nb_to_py_mapping.items():
             out, err = _replace_temp_python_file_references_in_out_err(
-                tmpdirname,
+                temp_python_file,
                 notebook,
                 out,
                 err,
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -69,44 +69,6 @@ def _temp_python_file_for_notebook(
     return temp_python_file
 
 
-def _replace_path_out_err(
-    out: str, err: str, temp_python_file: Path, notebook: Path
-) -> Tuple[str, str]:
-    """
-    Replace references to temporary Python file with notebook's path.
-
-    Parameters
-    ----------
-    out
-        Captured stdout from third-party tool.
-    err
-        Captured stderr from third-party tool.
-    temp_python_file
-        Temporary Python file where notebook was converted to.
-    notebook
-        Original Jupyter notebook.
-
-    Returns
-    -------
-    out
-        Stdout with temporary Python file replaced with notebook.
-    err
-        Stderr with temporary Python file replaced with notebook.
-    """
-    out = out.replace(str(temp_python_file), str(notebook))
-    err = err.replace(str(temp_python_file), str(notebook))
-
-    out = out.replace("   .py", ".ipynb")
-    err = err.replace("   .py", ".ipynb")
-
-    # This next part is necessary to handle cases when `resolve` changes the path.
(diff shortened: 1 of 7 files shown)


nbqa/__main__.py
```diff
@@ -201,8 +137,16 @@ def _replace_temp_python_file_references_in_out_err(
...
+    out = (
+        out.replace(f"{tempdirname}{os.sep}", "")
+        .replace(f"{str(Path(tempdirname).resolve())}{os.sep}", "")
+        .replace("   .py", ".ipynb")
+    )
+    err = (
+        err.replace(f"{tempdirname}{os.sep}", "")
+        .replace(f"{str(Path(tempdirname).resolve())}{os.sep}", "")
+        .replace("   .py", ".ipynb")
+    )
```
s-weigand, line 149: IMHO all that string replacing should stay function, which than can be unittested.

tests/test_other_magics.py
```diff
@@ -33,7 +33,7 @@ def test_cli(magic: str, expected: List[str], capsys: "CaptureFixture") -> None:
         main(["flake8", str(path), magic])
 
     out, _ = capsys.readouterr()
-    expected_out = "".join(f"{str(path.resolve())}:{i}{os.linesep}" for i in expected)
+    expected_out = "".join(f"{str(path)}:{i}{os.linesep}" for i in expected)
```
s-weigand, line 36: Since this is used 3 times this could be refactored to a function.

tests/test_black.py
```diff
@@ -31,7 +31,7 @@ def test_black_works(
     # check diff
     with open(tmp_notebook_for_testing) as handle:
         before = handle.readlines()
-    path = os.path.abspath(os.path.join("tests", "data", "notebook_for_testing.ipynb"))
+    path = os.path.join("tests", "data", "notebook_for_testing.ipynb")
```
s-weigand, line 34: Why not define a variable `TEST_DATA_PATH` at the root of tests and use this instead?
`TEST_DATA_PATH` could be of type `pathlib.Path` since it is only used in a strings, so no `str(path)` overhead.
That way this line (and all the similar ones) would be:
```python
path = TEST_DATA_PATH / "notebook_for_testing.ipynb"
```
Which IMHO is much more readable.

s-weigand (changes requested):

nbqa/__main__.py
s-weigand: But with only two calls, this is mostly aesthetics, so highly subjective 😝

MarcoGorelli: True :smile: IMO with just two simple calls, this is likely OK, but if we run into bugs down the line then extracting a function from here, with unit tests, might be in order

tests/test_black.py
MarcoGorelli: 100% agreed :+1: I'll do some much-needed cleaning up / refactoring here in a separate PR (I prefer to not mix changing logic with refactoring)

girip11 (changes requested):

nbqa/__main__.py
MarcoGorelli: @girip11 not sure I understand, sorry.

On `master`, I get:
```python
$ nbqa flake8 tests/data/notebook_for_testing.ipynb 
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_1:1:1: F401 'os' imported but unused
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_1:3:1: F401 'glob' imported but unused
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_1:5:1: F401 'nbqa' imported but unused
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_2:2:1: E302 expected 2 blank lines, found 0
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_2:19:9: W291 trailing whitespace
```

while on this branch I get

```python
$ nbqa flake8 tests/data/notebook_for_testing.ipynb 
tests/data/notebook_for_testing.ipynb:cell_1:1:1: F401 'os' imported but unused
tests/data/notebook_for_testing.ipynb:cell_1:3:1: F401 'glob' imported but unused
tests/data/notebook_for_testing.ipynb:cell_1:5:1: F401 'nbqa' imported but unused
tests/data/notebook_for_testing.ipynb:cell_2:2:1: E302 expected 2 blank lines, found 0
tests/data/notebook_for_testing.ipynb:cell_2:19:9: W291 trailing whitespace
```

If I just run `flake8`, I get

```
$ flake8 nbqa/replace_source.py --max-line-length=80
nbqa/replace_source.py:2:81: E501 line too long (85 > 80 characters)
nbqa/replace_source.py:43:81: E501 line too long (81 > 80 characters)
nbqa/replace_source.py:51:81: E501 line too long (85 > 80 characters)
nbqa/replace_source.py:70:81: E501 line too long (86 > 80 characters)
```

so wouldn't the change proposed in this PR would make the output more consistent with flake8's? (similarly for other tools)

girip11: Sorry. My bad. Ignore the comment.

girip11 (approved):

nbqa/__main__.py
MarcoGorelli: Your comment was valid though, as with that change we had

```python
$ nbqa flake8 /home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb
tests/data/notebook_for_testing.ipynb:cell_1:1:1: F401 'os' imported but unused
tests/data/notebook_for_testing.ipynb:cell_1:3:1: F401 'glob' imported but unused
tests/data/notebook_for_testing.ipynb:cell_1:5:1: F401 'nbqa' imported but unused
tests/data/notebook_for_testing.ipynb:cell_2:2:1: E302 expected 2 blank lines, found 0
tests/data/notebook_for_testing.ipynb:cell_2:19:9: W291 trailing whitespace
```

With the change I've just pushed, instead, we have:

```python
$ nbqa flake8 /home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_1:1:1: F401 'os' imported but unused
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_1:3:1: F401 'glob' imported but unused
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_1:5:1: F401 'nbqa' imported but unused
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_2:2:1: E302 expected 2 blank lines, found 0
/home/marco/nbQA-dev/tests/data/notebook_for_testing.ipynb:cell_2:19:9: W291 trailing whitespace

$ nbqa flake8 tests/data/notebook_for_testing.ipynb
tests/data/notebook_for_testing.ipynb:cell_1:1:1: F401 'os' imported but unused
tests/data/notebook_for_testing.ipynb:cell_1:3:1: F401 'glob' imported but unused
tests/data/notebook_for_testing.ipynb:cell_1:5:1: F401 'nbqa' imported but unused
tests/data/notebook_for_testing.ipynb:cell_2:2:1: E302 expected 2 blank lines, found 0
tests/data/notebook_for_testing.ipynb:cell_2:19:9: W291 trailing whitespace
```

which seems more like what we're after

I'll write a failing test (should've done that first :flushed:) and'll push again later

MarcoGorelli:
Seems to be fine now, tests contains a mix of absolute and relative paths and pass on both Windows and Linux :tada:

MarcoGorelli:
It seems `isort` resolves this differently on Windows (where it prints a full path) and on Linux, where it prints when you pass in.

So for consistency, what I've done in the last commit is to always print what the user passes in, regardless of what the tool would do.

If a user passes
```
nbqa <command> src/file.ipynb
```
the output will reference `src/file.ipynb`, while if the user passes
```
nbqa <command> /home/user/src/file.ipynb
```
the output will reference `/home/user/src/file.ipynb`.

I know you've reviewed already but this is a bit of an API breaking change, so any further comments on the latest commit would be appreciated @nbQA-dev/nbqa-core

girip11:
Sure will review the latest changes

s-weigand:
> So for consistency, what I've done in the last commit is to always print what the user passes in, regardless of what the tool would do.

I totaly agree, with the output being tool agnostic.
How about adding an option `path-format` = `absolute|relative`?
Since most users will call tools from the repo root, I would suggest `relative` as default.

MarcoGorelli:
That's just a good suggestion...I'm inclined to go with the `black` philosophy of not making the tool too configurable, mainly because like this it's easier to maintain

s-weigand:
I think one of the main points behind `back`s philosophy is.
> Black is the uncompromising Python code formatter.

So making it too confiurable, you end up with different code styles again, which kinda beats the purpouse.

Where as for a tool wrapper like `nbQA`, IMHO it would be preferable to have a uniform output no matter the tool used.

MarcoGorelli:
True, but with this PR it will be uniform - if the users wants the full path, they can pass a full path; if they want a relative path, they can pass a relative one.

I'm just not sure that having an extra flag for this would be worth the extra complexity / maintenance

s-weigand:
Ok point taken ^^

s-weigand:
Btw. isn't the replacing of the temp path making the argument behind using `   .py` for temp files invalid?
I meank, if a tool uses the console width for output formatting, the temppath replacement, will throw off the formatting much more than the 3 character difference wouldn't it?
Do you have an example tool that uses console width for formatting its output?

MarcoGorelli:
:thinking: ah, now that's a really good point...

> Do you have an example tool that uses console width for formatting its output?

Initially this was `pytest`, but then I realised it was _waaay_ to complex to support that so stopped. I guess now we can remove those extra whitespaces then, thanks :+1:

s-weigand (approved):

girip11 (approved):

PR #287, opened 2020-10-02
Credit Girish as co-author

MarcoGorelli:
Gotta give credit where it's due

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,8 +14,8 @@ setup_requirements = []
 test_requirements = []
 
 setup(
-    author="Marco Gorelli",
-    author_email="m.e.gorelli@gmail.com",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -14,8 +14,8 @@
 test_requirements = []
 
 setup(
-    author="Marco Gorelli",
-    author_email="m.e.gorelli@gmail.com",
+    author="Marco Gorelli, Girish Pasupathy",
+    author_email="",
```
s-weigand, line 18: Since `""` is the default for `author_email`, I guess you can just drop the line 😉 .
Or if @girip11 is fine with it add both your emails, if that is wanted.

girip11 (approved):

PR #288, opened 2020-10-03
Politely ask users to report bugs

MarcoGorelli:
Ask users to report a bug if the notebook fails to reconstruct:

```python-traceback
$ nbqa remove_comments tests/data/notebook_for_testing.ipynb --nbqa-mutate
  File "/home/marco/nbQA-dev/nbqa/__main__.py", line 495, in _run_on_one_root_dir
    replace_source.main(
  File "/home/marco/nbQA-dev/nbqa/replace_source.py", line 78, in main
    next(pycells),
StopIteration

The above exception was the direct cause of the following exception:

  File "/home/marco/nbQA-dev/.venv/bin/nbqa", line 33, in <module>
    sys.exit(load_entry_point('nbqa', 'console_scripts', 'nbqa')())
  File "/home/marco/nbQA-dev/nbqa/__main__.py", line 530, in main
    output_codes = [
  File "/home/marco/nbQA-dev/nbqa/__main__.py", line 531, in <listcomp>
    _run_on_one_root_dir(i, cli_args, configs, project_root)
  File "/home/marco/nbQA-dev/nbqa/__main__.py", line 499, in _run_on_one_root_dir
    raise RuntimeError(
RuntimeError: 

😭 Error reconstructing tests/data/notebook_for_testing.ipynb 😭

Please report a bug at https://github.com/nbQA-dev/nbQA/issues 🙏
```


neomatrix369:
Rule #1 write code without bugs
Rule #2 when unclear refer to Rule #1
Rule #3 enjoy bug-free code
😆 🤣

s-weigand:
How about using custom exceptions?

```python
class NbqaBaseException(Exception):
    def __init__(self, custom_msg: str, *args, **kwargs):

        ask_for_report_msg = (
            "Please report a bug at https://github.com/nbQA-dev/nbQA/issues 🙏"
        )
        super().__init__(f"{custom_msg}\n\n{ask_for_report_msg}", *args, **kwargs)


class NbqaPathReconstructError(NbqaBaseException, RuntimeError):
    def __init__(self, notebook: str, *args, **kwargs):
        reconstruct_error_msg = f"😭 Error reconstructing {str(notebook)} 😭"
        super().__init__(reconstruct_error_msg, *args, **kwargs)
```

MarcoGorelli:
Yes, that's a good idea...I'll check for error in other places too then

MarcoGorelli:
For now I've made a base error message, and will raise for unexpected errors while reconstructing or when parsing. Hopefully this can help us make the library better.

ATM I'm not sure I see the benefit of making custom errors and managing the extra classes that would come along with them, but I'm open to them being contributed in a separate PR if anyone feels like it

remove_comments.py
```diff
@@ -0,0 +1,17 @@
+"""
+Silly little module which removes comments.
+
+This is just so we can check what happens when running nbqa on a tool which causes a failure.
+"""
+
+import argparse
```
girip11, line 7: Should this module be placed under the tests directory, since this is only used for testing purposes and does not serve any use to the end user. We can add `tests/remove_comments.py` to PYTHONPATH before launching the tests and clean up after the test runs?

In `__main__.py`, may be we need to change `env["PYTHONPATH"] = os.getcwd()` to `env["PYTHONPATH"] = f"{env["PYTHONPATH"]}{os.pathsep}{os.getcwd()}`

girip11: Do you see any issues if we append to existing values of `PYTHONPATH`?

girip11 (changes requested):

remove_comments.py
MarcoGorelli: That's a good point, best to not clutter the root folder too much. I've made a fixture which makes a temporary copy.

> In __main__.py, may be we need to change env["PYTHONPATH"] = os.getcwd() to env["PYTHONPATH"] = f"{env["PYTHONPATH"]}{os.pathsep}{os.getcwd()}

Yes, this seems like a good idea...I'll make a separate PR for that

girip11: Ah! I like the fixture idea!

girip11 (approved):

PR #297, opened 2020-10-05
Prevent pre-commit autoupdate workflow to run on forks

s-weigand:
Using github.repository_owner instead of github.repository, allows using the workflow in other repositories of the same organization/user, while preventing it form running on forks.


MarcoGorelli:
Thanks! 

Here the organisation is `nbQA-dev` though, do we need to use that?

s-weigand:
Ups you are right 😮

PR #308, opened 2020-10-12
Remove trailing semicolon before subprocessrun

### LINKED ISSUE #305: If a cell ends with ; , and the rest of the notebook obeys the black formatter, then black will keep outputting 1
e.g. 

```python
plt.plot();
```

`nbqa black notebook.ipynb` will keep trying to remove the trailing semicolon, nbqa will keep reinstating it, and `nbqa black` will never return 0

MarcoGorelli:
closes #305 

The solution in this PR is to remove trailing semicolons _before_ saving the temporary Python file - we store where they are anyway so they'll get restored by `replace_source`, but like this we won't get false positives like those described in #305

### DIFF
--- a/nbqa/save_source.py
+++ b/nbqa/save_source.py
@@ -222,7 +222,7 @@ def main(
             )
             if parsed_cell.rstrip().endswith(";"):
                 trailing_semicolons.add(cell_number)
-            result.append(re.sub(r";(\s*)", "\\1", parsed_cell))
+            result.append(re.sub(r";(\s*)$", "\\1", parsed_cell))
             line_number += len(split_parsed_cell)
 
     with open(str(temp_python_file), "w") as handle:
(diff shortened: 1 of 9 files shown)


MarcoGorelli:
Have checked this on PyMC3's notebooks and the output looks good, haven't noticed any regressions - merging as I feel a bit uncomfortable having a buggy release, especially as new users may be finding the library (see Daniel-Mietchen/ideas#1347)

nbqa/save_source.py
```diff
@@ -222,6 +222,7 @@ def main(
             )
             if parsed_cell.rstrip().endswith(";"):
                 trailing_semicolons.add(cell_number)
+            result.append(re.sub(r";(\s*)$", "\\1", parsed_cell))
```
girip11, line 225: Nice!

tests/test_return_code.py
```diff
@@ -41,6 +44,13 @@ def test_black_return_code() -> None:
     expected = 0
     assert result == expected
 
+    output = subprocess.run(
```
girip11, line 47: We could use clean notebooks as a list and dirty notebooks as another list.

PR #310, opened 2020-10-12
Special-case isort

### LINKED ISSUE #293: Running nbqa with flake8 complains about blank lines on notebook cells with functions
Consider a cell containing a function

```Python
# cell contents
def get_dataset():
    return make_classification(
        n_samples=1000, n_features=20, n_informative=15, n_redundant=5, random_state=7
    )
```

I am getting the warnings 

* `E302 expected 2 blank lines, found 1`.
* `E305 expected 2 blank lines after class or function definition, found 1`

### LINKED ISSUE #309: If there is a cell after a def, nbqa black will report reformatting though nothing will change
Example

```python
# cell 1
def add(num1, num2):
    return num1 + num2

# cell 2
add(2, 3)

# cell3

```

We then get
```bash
$ python -m nbqa black tests/data/notebook_with_cell_after_def.ipynb --nbqa-mutate
reformatted tests/data/notebook_with_cell_after_def.ipynb
All done! ✨ 🍰 ✨
1 file reformatted.
$ git status
On branch cell-after-def
nothing to commit, working tree clean
```

MarcoGorelli:
closes #309 
closes #293 

I couldn't think of a better way to do this than to special-case isort.

The idea is that generally the cell separator will be `f"\n\n{CODE_SEPARATOR}\n"`, except for `isort` when it'll be `f"\n{CODE_SEPARATOR}\n"`

@girip11 you'd suggested this in https://github.com/nbQA-dev/nbQA/issues/293#issuecomment-703712225 , 

> Adding two new lines seems to solve the problem

Special-casing isort to just have a single newline would solve the flake8 bug, and this new black bug.

I'm not a fan of special-casing, but...I have a feeling this might be the only way here

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -427,7 +427,10 @@ def _run_on_one_root_dir(
         for notebook, temp_python_file in nb_to_py_mapping.items():
             try:
                 nb_info_mapping[notebook] = save_source.main(
-                    notebook, temp_python_file, configs.nbqa_ignore_cells
+                    notebook,
+                    temp_python_file,
+                    configs.nbqa_ignore_cells,
+                    cli_args.command,
                 )
             except Exception as exc:
                 raise RuntimeError(
--- a/nbqa/save_source.py
+++ b/nbqa/save_source.py
@@ -219,6 +219,7 @@ def main(
             parsed_cell = _parse_cell(
                 cell["source"], cell_number, temporary_lines, command
             )
+            result.append(parsed_cell)
             cell_mapping.update(
                 {
                     j + line_number + 1: f"cell_{cell_number}:{j}"
--- a/nbqa/save_source.py
+++ b/nbqa/save_source.py
@@ -18,6 +18,8 @@ if TYPE_CHECKING:
 
 CODE_SEPARATOR = "# %%"
 MAGIC = ["%%script", "%%bash"]
+NEWLINES = defaultdict(lambda: "\n\n")
+NEWLINES["isort"] = "\n"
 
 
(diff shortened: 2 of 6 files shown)


MarcoGorelli:
Have checked this on PyMC3's notebooks and the output looks good, haven't noticed any regressions - merging as I feel a bit uncomfortable having a buggy release, especially as new users may be finding the library (see Daniel-Mietchen/ideas#1347)

s-weigand:
You can yank releases from PyPi, that way they can only be installed if the specific version is specified.
Also, it can only be seen on PyPi, if you look at the history.
Basically that is the user friendly way of deleting a release 😉

MarcoGorelli:
I didn't know that (as you can no doubt tell, this is my first time releasing my own package) - I'll do that next time then, thanks!

s-weigand:
It's a quite new feature (1y?), but very useful if you messed something up in release, i.e. made the package py36+ and forgot to set the proper `python_requires`. I didn't do that myself (yet!), but had the issue a few times with dependencies dropping py27 support w/o setting `python_requires` and back then you would have to delete all uploaded files and reupload the new ones, which ofc no one wanted to do 😋

girip11:
> I'm not a fan of special-casing, but...I have a feeling this might be the only way here

I too would have gone with this approach. 

In case of isort, if no imports are to be sorted, then it can just leave the file untouched and leave the formatting to formatting tools like black, yapf. I am not sure if this behavior can be called a bug with isort.

MarcoGorelli:
> I am not sure if this behavior can be called a bug with isort.

Maybe, but given how popular they are, I think they'll be unlikely to change this behaviour

PR #312, opened 2020-10-12
Allow nbqa to modify notebooks when run on directory

### LINKED ISSUE #311: nbqa doesn't modify notebooks when run on directory
It's because we still have

```
def _get_mtimes(arg: Path) -> Set[float]:
    """
    Get the modification times of any converted notebooks.

    Parameters
    ----------
    arg
        Notebook or directory to run 3rd party tool on.

    Returns
    -------
    Set
        Modification times of any converted notebooks.
    """
    if not arg.is_dir():
        return {os.path.getmtime(str(arg))}
    return {os.path.getmtime(str(i)) for i in arg.rglob("*   .py")}

```

...my fault of hardcoding `   .py`

MarcoGorelli:
closes #311

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -347,6 +347,7 @@ def _run_command(
         cwd=tmpdirname,
         env=env,
     )
+
     mutated = _get_mtimes(arg) != before
 
     output_code = output.returncode
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -265,10 +265,10 @@ def _get_arg(
     >>> root_dir = "root_dir"
     >>> tmpdirname = "tmpdir"
     >>> nb_to_py_mapping = {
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
Have checked this on PyMC3's notebooks and the output looks good, haven't noticed any regressions - merging as I feel a bit uncomfortable having a buggy release, especially as new users may be finding the library (see https://github.com/Daniel-Mietchen/ideas/issues/1347)

girip11:
In the test case added (or in a separate test case), we could also check the modified times of the notebooks in addition to checking a diff on a single file. 

I can send a PR for that today.

MarcoGorelli:
If you want to, yes, that would be a better test!

To clean up the tests, we could also use this fixture consistently and get rid of the others (instead of `tmp_notebook_for_testing`, just use `tmp_test_data/'notebook_for_testing.ipynb'`)

PR #317, opened 2020-10-13
Tests refactoring and cleanup

girip11:
Changes in this PR

1. Moved all the third party tool tests to `tests/tools` directory. It's easier to find the tool if its in a single directory
2. Merged all tests cases throwing runtime errors like missing_command, missing_root, etc that were in spread in different files.
3. Refactored test cases related to magics
4. Refactored `test_return_code` module
5. Added extra check to `test_version` test case to test the version.


MarcoGorelli:
Awesome work!! Your cleanups are really impressive, the codebase looks so much better since you started contributing! I felt that the complexity of the tests was increasing, so this PR is very much appreciated. It's quite big so I'll review/merge tomorrow

tests/test_ipython_magics.py
```diff
@@ -0,0 +1,194 @@
...
+
+if TYPE_CHECKING:
+    from _pytest.capture import CaptureFixture
+    from py._path.local import LocalPath
+
+
+def _create_ignore_cell_config(config_file_path: Path, config: str) -> None:
+    """Create configuration file for nbqa."""
+    with open(str(config_file_path), "w") as handle:
+        handle.write(config)
```
MarcoGorelli, line 20: while we're here, this could be `.write_text`?

tests/test_return_code.py
```diff
@@ -2,69 +2,51 @@
 
 import os
 import subprocess
+from functools import partial
+from typing import List
 
 DIRTY_NOTEBOOK = os.path.join("tests", "data", "notebook_for_testing.ipynb")
 CLEAN_NOTEBOOK = os.path.join("tests", "data", "clean_notebook.ipynb")
```
s-weigand, line 9: This could be rewritten with `Path` and a `TEST_DATA_DIR` variable to make it more concise.

tests/test_return_code.py
```diff
@@ -1,70 +1,55 @@
...
-    expected = 1
-    assert result == expected
+    clean_notebooks = [
+        CLEAN_NOTEBOOK,
+        TEST_DATA_DIR / "notebook_with_cell_after_def.ipynb",
+        TEST_DATA_DIR / "clean_notebook_with_trailing_semicolon.ipynb",
+    ]
+    assert black_runner(clean_notebooks, "--check") == PASSED
 
-    output = subprocess.run(["nbqa", "black", CLEAN_NOTEBOOK, "--check", "-l", "1"])
```
MarcoGorelli, line 67: any chance this test (with `-l 1`) could be preserved? I can't remember why it was added, but just in case I'd rather keep it

tests/test_version.py
```diff
@@ -1,9 +1,15 @@
...
+from nbqa import __version__
 
 
 def test_version() -> None:
     """Check you can run :code:`nbqa --version`."""
-    output = subprocess.run(["nbqa", "--version"])
+    output = subprocess.run(
+        ["nbqa", "--version"], stdout=PIPE, stderr=PIPE, universal_newlines=True
+    )
+    assert output.stdout.strip() == f"nbqa {__version__}"
```
MarcoGorelli, line 14: nice! :+1:

tests/test_ipython_magics.py
```diff
@@ -0,0 +1,193 @@
...
+        result = result and (actual_src == expected_src)
+
+    return result and (actual == "")
+
+
+@pytest.mark.parametrize(
+    "config, validate, config_file",
+    [*_indented_magics_toml_input()],
+)
+def test_indented_magics(
```
MarcoGorelli, line 170: nice cleanup here!

MarcoGorelli (changes requested):
Thanks for updating - I just have a small comment about one test which I think it would be good to preserve, feel free to merge (or ping us and we can do it) if you agree / if that's done

tests/test_return_code.py
girip11: I wanted to add this test case to `test_black.py`. I wanted the test case to check the contents of the formatted file not just the return code. May be I should keep this and remove it when I add the test case to black. That would be better.

tests/test_ipython_magics.py
girip11: There is still some code duplication(code between the test case functions are similar). But I didnt want to go to extreme refactoring. If we are finding a pattern out of this, then we can further refactor.

tests/test_return_code.py
MarcoGorelli: >  I wanted the test case to check the contents of the formatted file not just the return code

That's a better idea actually :+1: let's take this PR as is and then that can be a separate one

PR #338, opened 2020-10-17
Adding Sourcery-ai badge to the README

neomatrix369:
Minor cosmetic change to the README


MarcoGorelli:
Thanks @neomatrix369 - TBH I'd rather not clutter up the readme too much, I'd rather just have it focused on things users should know (what versions of Python are supported, that it supports pre-commit, that we have good code coverage) - if we put a badge for every tool we use then it'll be too cluttered. It was like that at the start but we then got rid of many of them in #291

neomatrix369:
Ok sounds good, also pre-commit didn't like it :D

PR #340, opened 2020-10-17
use pathlib

### LINKED ISSUE #240: CLN use Path.write_text where possible

MarcoGorelli:
closes #240

### DIFF
--- a/nbqa/replace_source.py
+++ b/nbqa/replace_source.py
@@ -85,8 +85,7 @@ def main(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) -> 
     notebook_info
         Information about notebook cells used for processing
     """
-    with open(notebook) as handle:
-        notebook_json = json.load(handle)
+    notebook_json = json.loads(notebook.read_text())
 
     pyfile = python_file.read_text()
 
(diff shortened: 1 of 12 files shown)


nbqa/replace_source.py
```diff
@@ -88,8 +88,7 @@ def main(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) ->
     with open(notebook) as handle:
         notebook_json = json.load(handle)
```
girip11, line 89: You could also write this using `json.loads(notebook.read_text())`

tests/test_configs_work.py
```diff
@@ -23,8 +23,7 @@ def test_configs_work(capsys: "CaptureFixture") -> None:
     capsys
         Pytest fixture to capture stdout and stderr.
     """
-    with open(".flake8", "w") as handle:
-        handle.write("[flake8]\nignore=F401\nselect=E303\nquiet=1\n")
+    Path(".flake8").write_text("[flake8]\nignore=F401\nselect=E303\nquiet=1\n")
```
girip11, line 26: How about using dedent for this too?

girip11 (changes requested):

nbqa/replace_source.py
MarcoGorelli: nice!

MarcoGorelli: seems there's a _slight_ performance hit from this, but it's probably fine for the sake of readability
```python
In [8]: %timeit json.loads(Path(notebook).read_text())                                                                                                                                                    
41 µs ± 159 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [9]: %timeit with open(notebook) as handle: json.load(handle)                                                                                                                                          
31.7 µs ± 129 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

girip11 (approved):

Issue #321, opened 2020-10-15
Provision for user to troubleshoot `Command not found.` error


girip11:
Currently, if nbqa and the QA tool to run are installed in different python environments, we would get `ValueError: Command black not found. Please make sure you have it installed before running nbqa on it.`

This feature adds a provision in nbqa to dump the path of the tool that will be run by nbqa using a command `which` or a flag `--which`. This can be print user friendly message if the tool cannot be found by nbqa. 

It could be something like 
* `nbqa which black` - This will introduce a new command.

This would make the user to better troubleshoot this issue and we can document how to handle this scenario in nbqa's documentation, thus making the user solve such issues on their own.

MarcoGorelli:
Yes, that's probably better...if a user gets `command not found`, I think they'll likely give up on using this library if they can't find a solution before too long, so collecting all the necessary information for debugging right away is probably best

s-weigand:
With #325 it is required that the tool is installed in the same env as `nbqa`.
So the check if the tool is installed can be done with `from importlib.util import find_spec`.

s-weigand:
Also `nbqa` could list the installed and supported tools with the `--version` flag as `flake8` does.
```
$ flake8 --version
3.8.3 (mccabe: 0.6.1, pycodestyle: 2.6.0, pyflakes: 2.2.0) CPython 3.7.8 on Windows
```

girip11:
This is the approach I am thinking of

* Introduce an option `nbqa black --find` to display black location, version etc as found by nbqa. I can use `importutil`, `imp`, `pkg_resources` modules. This is the success scenario where nbqa is able to find the tool.

* If nbqa can't find black, then we will display appropriate diagnostic message like what is nbqa installation path, and the used python executable path, then asking users to either install `pip(from same location as sys.executable) install black` or point to setting up python virtual environment. If case where user can't resolve this issue, then we can display a message to file bug/issue at nbqa github repo.

* When the user runs `nbqa black abc.ipynb --nbqa-mutate`, if the nbqa raises `ValueError("Command not found")` in that exception message in addition to what we display today, we will also ask the user to run `nbqa black --find` to troubleshoot this issue.

This one can be done with minimal changes and it will serve the purpose I think. Please reply if you have any queries.

s-weigand:
Instead of looking for `ValueError("Command not found")` with speciffic message, a custom error (e.g. `NbqaMissingCommandError`) might be a cleaner approach.

[...]

PR #356, opened 2020-10-20
Add `--nbqa-find` to help the users to troubleshoot issues where nbqa…

girip11:
… is not able to run the command and raises `ValueError: Command not found.`

Changes to be committed:
	modified:   nbqa/__main__.py
	modified:   nbqa/cmdline.py
	new file:   nbqa/troubleshoot.py
	modified:   tests/test_runtime_errors.py
	new file:   tests/test_troubleshoot.py


girip11:
This change addresses https://github.com/nbQA-dev/nbQA/issues/321. I have introduced `--nbqa-find` flag to check if nbqa can locate the command.

Currently the command looks like `nbqa flake8 --nbqa-find .` . Notice we still need to provide a notebook path or directory for this command to run. I tried to get rid of passing atleast one path to the command. I tried creating using `argument_parser.add_mutually_exclusive_group()`, but no success.

I thought `nbqa flake8 --nbqa-find` looks better compared to `nbqa flake8 --nbqa-find .`.

MarcoGorelli:
would `nbqa --nbqa-find flake8` be possible? I ran into problems in the past when trying to let root_dirs be optional

girip11:
But making `nargs="*"` did not work and also that might not be correct too.

girip11:
I have made this change ready for review. If we feel this solution is kind of ugly, I can close this PR.

MarcoGorelli:
Thanks for looking into this.

Why not just print a few more diagnostics in the
```
Command `{command}` not found.
            Please make sure you have it installed in the same environment as nbqa.
```
error message, such as the path of nbqa, python, and the command? This way we don't need to add a new flag, if it's not possible to do it in a way which we feel good about?

girip11:
> Thanks for looking into this.
> 
> Why not just print a few more diagnostics in the
> 
> ```
> Command `{command}` not found.
>             Please make sure you have it installed in the same environment as nbqa.
> ```
> 
> error message, such as the path of nbqa, python, and the command? This way we don't need to add a new flag, if it's not possible to do it in a way which we feel good about?

I thought of that initially and ruled out in favor of keeping the error message less verbose. I even commented the same in the issue.  But now that seems to be better approach and simpler too.

```Python
if "No module named" in err:
        # Collect path of python that will be used by nbqa
        # nbqa version used
        # PATH env variable as seen by nbqa
        # Add these collected information to the ValueError message
        raise ValueError(
            f"Command `{command}` not found. "
            "Please make sure you have it installed before running nbqa on it."
        )
```

MarcoGorelli:
> I even commented the same in the issue. But now that seems to be better approach and simpler too.

girip11:
So I will make the changes and send next iteration.

s-weigand:
While we should have useful debug messages, IMHO we don't need to force users to install the same tools in each venv (e.g. one global `black` is sufficient). Also, many users install all packages in a global fashion (e.g. `conda base`).
We can call any tool that is found by the shell, with using `subprocess.run("command", shell=True)`

Here is a small dummy example:
```console
$ python -m venv venv1
$ source venv1/bin/activate
$ which python
/mnt/d/git/del_me/dummy/venv1/bin/python
$ which flake8
/home/dafuq/anaconda3/bin/flake8
```

So we got the venv python and `flake8` installed globally.
Since `flake8` isn't installed in the venv, this command fails:
```console
$ python -c 'import subprocess; subprocess.run("python -m flake8 --help")'
  File "<string>", line 1, in <module>
  File "/home/dafuq/anaconda3/lib/python3.8/subprocess.py", line 489, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/home/dafuq/anaconda3/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/home/dafuq/anaconda3/lib/python3.8/subprocess.py", line 1702, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'python -m flake8 --help'
```

But if we use `shell=True`, this happens:
```console
python -c 'import subprocess; subprocess.run("flake8 --help", shell=True)'
usage: flake8 [options] file file ...
... <rest of the help text>
```

MarcoGorelli:
@s-weigand that's a good point, I just wonder if we'd run into issues whereby we run the system installation of a tool, while `python -m <command>` would've run the one in the venv?

Not a blocker, but we'd need to special-case doctest, as that needs to be run as `python -m doctest`.

I see having separate environments as a good practice anyway, and possibly something I'd like to encourage anyway.

So, I'm not sure what I'd like most here, I need to do some more reading / thinking / experimenting, but thanks for having brought it up :+1:

MarcoGorelli:
From https://docs.python.org/3/library/subprocess.html :

> If shell is True, the specified command will be executed through the shell. This can be useful if you are using Python primarily for the enhanced control flow it offers over most system shells and still want convenient access to other shell features such as shell pipes, filename wildcards, environment variable expansion, and expansion of ~ to a user’s home directory

> When using shell=True, the shlex.quote() function can be used to properly escape whitespace and shell metacharacters in strings that are going to be used to construct shell commands.

OK, maybe it's fine then - we can use `shlex.quote`, and then, seeing as users would be running this on their own machines anyway, security is arguably less of an issue?

s-weigand:
As far as I know, all venv implementations modify the PATH by prepending the env path (first found gets used).

A quick test with pip shows that the right version gets used in a venv.
```console
$ python -c 'import subprocess; subprocess.run("which pip", shell=True)'
/mnt/d/git/del_me/dummy/venv1/bin/pip
```

Using `shell=True`, would also allow using none python tools.
E.g. I could think of wanting to use `grep` on notebooks, to find links that have changed (e.g. the `master` -> `main` renaming on github) and have a per cell formatting (not sure if it would work out of the box 😋 ).

`shlex.quote` takes care of shell injection (shell pondon for SQL injections), which IMHO is mainly a problem if you use it in a web context, where you don't want a user to be able to execute arbitrary shell commands on your server.
In the case of `nbqa` the user has already access to the shell, so I don't see that issue here (why would you want to do `nbqa flake8; sudo rm -rf /` when you can just do `sudo rm -rf /` 😅 ).

In case someone wants to use it in a web context (e.g. jupyter extension + `jupyterhub`) a `--no-shell` flag could be added.

But I see the main usage in local development and a CI context, where IMHO more freedom is mostlikely more desired by users.
Thus I think 'use tooling you would use on *.py files on *.ipynb files' is what I would want as a user (e.g. `grep`).

MarcoGorelli:
OK, those are some really good points! Thanks for having laid them out so clearly. @girip11 , thoughts?

I'm happy to go with shell=True, escaping using shlex, and special-casing doctest to run with `sys.executable -m`

Nonetheless, I feel like that belongs in a separate PR - these diagnostics may still be useful!

s-weigand:
Totally agree
> Nonetheless, I feel like that belongs in a separate PR - these diagnostics may still be useful!

girip11:
> Using shell=True, would also allow using none python tools.
> E.g. I could think of wanting to use grep on notebooks, to find links that have changed (e.g. the master -> main renaming on github) and have a per cell formatting (not sure if it would work out of the box yum ).

Checkout `jupytext`.

Right now, I would like nbQA to be just a tool to run QA tools on notebooks. We could use tools like `jupytext` for scenarios like running a grep. `jupytext` has the capability to convert the notebook to various formats including to `.py` file. `jupytext  a.ipynb --to py - | grep <something>`. In my opinion, that would offer more flexibility. 

The reason why nbQA is not suitable right now to run any tool, is nbqa modifies the source notebook replacing all the magics etc. Basically nbqa does a whole lot of preprocessing on the notebooks before applying the tools. Thus nbQA atleast for now should not diverge from what it is written for.

> Also, many users install all packages in a global fashion (e.g. conda base).

If that's the case, then those users would most probably have installed `nbqa` at the same location because it is always better to group similar purpose tools at one single location(may be all dev dependencies can be installed globally instead of having it in each venv). If not haven't, then its better to do so.

> I see having separate environments as a good practice anyway, and possibly something I'd like to encourage anyway.

I agree with @MarcoGorelli  on this. 

* So my conclusion regarding the `shell=True`, let users report issues on such kinds of usability issues and if there are many such complaints then we can seriously consider it. This will be an issue only in local development right? 

* Because I don't see folks installing the packages at different locations in CI. 

* Also there would not be any issue in using nbqa as pre-commit hook because additional dependencies take care installing those in a isolated environment(with the latest changes).

girip11:
> So we got the venv python and flake8 installed globally.
Since flake8 isn't installed in the venv, this command fails:

@s-weigand  This should not matter now right? Because if we are using  `sys.executable` then python is picked relative to that of `nbqa` because it uses `argv[0]` which is the script name in this case. Reason is if `nbqa` and flake8 are installed globally even if we have activated `python venv`, sys.executable called from nbqa will pick up global python executable right?

MarcoGorelli:
> This will be an issue only in local development right?

Indeed, if we encourage this tool to be run via pre-commit (which IMO is the best way to run virtually any linter/formatter) then it won't be an issue. I'll try to make that a bit more explicit in the docs/readme

> let users report issues on such kinds of usability issues and if there are many such complaints then we can seriously consider it

OK, let's go with this for now - most users don't report bugs, but still, it seems unlikely that someone would have a global installation of flake8 but a _local_ installation of nbqa. Encouraging consistency on their side should help reduce surprises later on, which reduces the chances of them blaming our tool when something goes wrong

s-weigand:
@girip11 You are right, there are no problems when run in the CI or via pre-commit.
The only problematic case is when run from a venv, while the tool is only installed globally.
But users who install all packages globally will also have `nbqa` globally (kinda missed that 😅 )

girip11:
@MarcoGorelli : This PR is ready for review. I addressed the comments.

MarcoGorelli:
Thanks @girip11 , looks great, and thanks for fixing my incorrect `ValueError` with `ModuleNotFoundError`!

PR #358, opened 2020-10-21
Add extras_require, to install nbqa with all supported tools

### LINKED ISSUE #357: Add extra_require with all supported tools
For users it might be convenient to install `nbqa` with all its supported tools, using a single command.
This can be done by adding the [`extra_require` kwarg to the setup function in `setup.py`](https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies).

E.g. if the `extra_require` is called `toolchain`, users can install it by running
```console
$ pip install nbqa[toolschain]
```

Which will install `nbqa` and all dependecies listed in `extra_require["toolchain"]`.

s-weigand:
This will allow users to install `nbqa` + all supported tools by running 
```console
$ python -m pip install -U nbqa[toolchain]
```

Maybe there is a better name than `toolchain` e.g. `full`, `all`, `tools` (maybe `+tools`, which has a nice readability)

closes #357

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ Install `nbqa` in your [virtual environment](https://realpython.com/python-virtu
 python -m pip install -U nbqa
 ```
 
+If you want to install `nbqa` together with all tools it supports, you can do so by running:
+
+```bash
+python -m pip install -U nbqa[toolchain]
+```
+
 ## 🚀 Examples
 
 Reformat your notebooks with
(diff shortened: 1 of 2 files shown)


girip11 (approved):

s-weigand:
Ok `+tools` gives a parsing error 😝

MarcoGorelli:
Thanks @s-weigand !

MarcoGorelli:
am temporarily removing this from the readme though, will add it in when we make the next release

MarcoGorelli:
I'm planning on doing a bit of an overhaul of the docs this weekend, so I'll figure out the best place to mention this then

Issue #355, opened 2020-10-19
ENH try parsing code after cell magic, ignore if can't


MarcoGorelli:
I haven't (yet) got to the root of the issue, but the notebook in question is  

https://github.com/microsoft/computervision-recipes/blob/1bb489af757fde7c773e16fab87b24305cff4457/scenarios/detection/20_deployment_on_kubernetes.ipynb

----

I get

```console
$ nbqa black scenarios/detection/20_deployment_on_kubernetes.ipynb
error: cannot format scenarios/detection/20_deployment_on_kubernetes.ipynb: Cannot parse: 50:1: [print(f) for f in Path(data_path()/"odFridgeObjects"/"models").glob('**/*')]
Oh no! 💥 💔 💥
1 file failed to reformat.
```

s-weigand:
For cases like this a `--keep-parsed-notebooks` flag is usefull (kinda related to #321)

MarcoGorelli:
Yes, that'd be good!

Even with that, I can't tell what the problem is:
```python
# %%
import sys
sys.path.append("../../")

import os
from PIL import Image
from pathlib import Path
import requests
import json
import inspect

from utils_cv.common.image import ims2strlist
from utils_cv.common.data import data_path
from utils_cv.detection.model import DetectionLearner, get_pretrained_fasterrcnn
from utils_cv.detection.dataset import DetectionDataset
from utils_cv.detection.plot import (
    display_bboxes,
    PlotSettings,
    DetectionBbox
)
from utils_cv.common.azureml import get_or_create_workspace

from azureml.core.webservice import AksWebservice
from azureml.core.compute import AksCompute, ComputeTarget
from azureml.core.model import Model, InferenceConfig


# %%
type(0xDD9F885E)  # %reload_ex 0xDD9F885E
type(0xA4B2F1E3)  # %autoreloa 0xA4B2F1E3
type(0xA34775A)  # %matplotli 0xA34775A


# %%
# Azure resources
subscription_id = "YOUR_SUBSCRIPTION_ID"
resource_group = "YOUR_RESOURCE_GROUP_NAME"  
workspace_name = "YOUR_WORKSPACE_NAME"  
workspace_region = "YOUR_WORKSPACE_REGION"

# Choose a size for our cluster and the maximum number of nodes
VM_SIZE = "STANDARD_NC6" #STANDARD_NC6S_V3"
NODES = 3
KUBERNETES_CLUSTER_NAME = "od-aks-cluster"
[...]

PR #361, opened 2020-10-21
hard-code cell magics to ignore

MarcoGorelli:
@girip11 with reference to #355

I see this as a temporary solution. Basically, on Friday, a new episode of the podcast Python Bytes will be out, and they said they'll mention our library.

My hope is that this will bring some new users here.

I don't think we can implement #355 in time, so my suggestion would be to, for now, hard-code all the magics from  https://ipython.readthedocs.io/en/stable/interactive/magics.html (so that any new user likely has a smooth user experience without having to learn about `--nbqa-ignore-cells`), and then once #355 is addressed properly we can deprecate `--nbqa-ignore-cells`.

@nbQA-dev/nbqa-core thoughts? Any objections?

----

I haven't added extra test cases as coverage isn't affected and we already have tests to check that cell magics listed in this constant are indeed ignored

### DIFF
--- a/nbqa/save_source.py
+++ b/nbqa/save_source.py
@@ -27,6 +27,7 @@ MAGIC = [
     "%%latex",
     "%%markdown",
     "%%perl",
+    "%%python2",
     "%%ruby",
     "%%script",
     "%%sh",
--- a/nbqa/save_source.py
+++ b/nbqa/save_source.py
@@ -18,7 +18,21 @@ if TYPE_CHECKING:
     from pathlib import Path
 
 CODE_SEPARATOR = "# %%"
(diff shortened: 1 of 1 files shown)


girip11:
These changes don't touch the package's public interface. So I am fine with this. As you said when the changes addressing #355 are ready, this can be fixed as well.

girip11:
I will work on #355 as the next item. But I won't do anything to release sooner to release it out because I wanted to do some thorough testing. So this change gives me the time I needed.

girip11 (approved):

MarcoGorelli:
> But I won't do anything to release sooner to release it out because I wanted to do some thorough testing

Awesome - please take your time, no hurry! :smile:

PR #372, opened 2020-10-24
Optimise for pre commit

### LINKED ISSUE #371: PERF only run third-party tool once
Rather than doing
```
    output_codes = [
        _run_on_one_root_dir(i, cli_args, configs, project_root)
        for i in cli_args.root_dirs
    ]
```
we can probably get better performance if we first get all the notebooks we need to convert, and then pass them in one-by-one

MarcoGorelli:
closes #371 

----

10x speedup!

```
(.venv) marco@marco-Predator-PH315-52:~/pymc3-dev$ time pre-commit run nbqa-isort --all
nbqa-isort...............................................................Passed

real    0m11.657s
user    0m10.401s
sys     0m1.252s
(.venv) marco@marco-Predator-PH315-52:~/pymc3-dev$ time pre-commit run local-nbqa-isort --all
local-nbqa-isort.........................................................Passed

real    0m0.987s
user    0m7.340s
sys     0m0.913s
```

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -296,7 +296,7 @@ def _get_all_args(
     tmpdirname: str,
     nb_to_py_mapping: Dict[Path, Path],
     project_root: Path,
-) -> Iterator[Path]:
+) -> List[Path]:
     """
     Get all arguments to run command against.
 
@@ -313,10 +313,10 @@ def _get_all_args(
 
     Returns
     -------
-    Path
+    List[Path]
         All notebooks or directories to run third-party tool against.
     """
-    return (_get_arg(i, tmpdirname, nb_to_py_mapping, project_root) for i in root_dirs)
+    return [_get_arg(i, tmpdirname, nb_to_py_mapping, project_root) for i in root_dirs]
 
 
 def _get_mtimes(arg: Path) -> Set[float]:
@@ -342,7 +342,7 @@ def _run_command(
     command: str,
     tmpdirname: str,
     cmd_args: List[str],
-    args: Iterator[Path],
+    args: List[Path],
 ) -> Tuple[str, str, int, bool]:
     """
     Run third-party tool against given file or directory.
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -29,7 +29,7 @@ BASE_ERROR_MESSAGE = dedent(
 )
 
 
-def _get_notebooks(root_dir: str) -> Iterator[Path]:
+def _get_notebooks(root_dir: List[str]) -> Iterator[Path]:
     """
     Get generator with all notebooks in directory.
 
@@ -48,6 +48,9 @@ def _get_notebooks(root_dir: str) -> Iterator[Path]:
     return (
         i for i in Path(root_dir).rglob("*.ipynb") if ".ipynb_checkpoints" not in str(i)
     )
+
+def _get_all_notebooks(root_dir):
+    return (j for i in root_dir for j in _get_notebooks(i))
 
 
 def _temp_python_file_for_notebook(
@@ -268,11 +271,14 @@ def _get_arg(
     'tmpdir/my_notebook.py'
     """
     if Path(root_dir).is_dir():
-        arg = Path(tmpdirname) / Path(root_dir).resolve().relative_to(project_root)
+        arg = [Path(tmpdirname) / Path(root_dir).resolve().relative_to(project_root)]
     else:
-        assert len(nb_to_py_mapping) == 1
-        arg = next(iter(nb_to_py_mapping.values()))
+        arg = list(nb_to_py_mapping.values())
     return arg
+
+def _get_all_args(root_dirs, tmpdirname, nb_to_py_mapping, project_root):
+    breakpoint()
+    return ' '.join(_get_arg(i, tmpdirname, nb_to_py_mapping, project_root) for i in root_dirs)
 
 
 def _get_mtimes(arg: Path) -> Set[float]:
@@ -335,17 +341,17 @@ def _run_command(
         "PYTHONPATH"
     ] = f"{env.get('PYTHONPATH', '').rstrip(os.pathsep)}{os.pathsep}{os.getcwd()}"
 
-    before = _get_mtimes(arg)
+    before = [_get_mtimes(i) for i in arg]
 
     output = subprocess.run(
-        [sys.executable, "-m", command, str(arg), *cmd_args],
+        [sys.executable, "-m", command, ' '.join(str(i) for i in arg), *cmd_args],
         stderr=subprocess.PIPE,
         stdout=subprocess.PIPE,
         cwd=tmpdirname,
         env=env,
     )
 
-    mutated = _get_mtimes(arg) != before
+    mutated = [_get_mtimes(i) for i in arg] != before
 
     output_code = output.returncode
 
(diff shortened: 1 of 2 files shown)


nbqa/__main__.py
```diff
@@ -270,11 +287,38 @@ def _get_arg(
...
+        Mapping between notebooks and Python files corresponding to converted notebooks.
+    project_root
+        Root of repository, where .git / .hg / .nbqa.ini file is.
+
+    Returns
+    -------
+    List[Path]
+        All notebooks or directories to run third-party tool against.
+    """
+    return [_get_arg(i, tmpdirname, nb_to_py_mapping, project_root) for i in root_dirs]
```
neomatrix369, line 319: +1 for using list comprehensions, adds to speedup

nbqa/__main__.py
```diff
@@ -298,7 +342,7 @@ def _run_command(
     command: str,
     tmpdirname: str,
     cmd_args: List[str],
-    arg: Path,
+    args: List[Path],
```
neomatrix369, line 345: minor hint: a list is slower/takes memory compared to tuples

MarcoGorelli: thanks! will change

nbqa/__main__.py
neomatrix369, line 319: Minor hint: using enumerators help use memory more efficiently and also helps with CPU usage eventually

nbqa/__main__.py
```diff
@@ -335,17 +379,17 @@ def _run_command(
...
-        [sys.executable, "-m", command, str(arg), *cmd_args],
+        [sys.executable, "-m", command, *(str(i) for i in args), *cmd_args],
         stderr=subprocess.PIPE,
         stdout=subprocess.PIPE,
         cwd=tmpdirname,
         env=env,
     )
 
-    mutated = _get_mtimes(arg) != before
+    mutated = [_get_mtimes(i) for i in args] != before
```
neomatrix369, line 392: same here 
```
mutated = [_get_mtimes(i) for _, i in enumerate(args)] ...
```

nbqa/__main__.py
MarcoGorelli: :thinking: are you sure? I don't see why using an unused enumerator would help

neomatrix369:
As hinted on gitter, when I get a moment I can do a module by module perf analysis on CPU and then on memory optimisation, but feel free to beat me to it. I'm always happy to comment on perf improvements.

nbqa/__main__.py
neomatrix369: Try to apply `enumerate` to other `for` loops as well

neomatrix369 (commented):
Overall looks good,
except you are not capturing before and after perf metrics and comparing them via separate perf tests like in this example here:  see https://github.com/neomatrix369/nlp_profiler/blob/master/slow-tests/performance_tests/test_perf_grammar_check.py (have a look at https://github.com/neomatrix369/nlp_profiler/blob/master/slow-tests/performance_tests/common_functions.py as well)

I usually fail tests that are off by 10% of the best benchmark I have gathered, as these tests are pretty non-deterministic - but anything beyond that leads me to investigate the issue

Use any other paradigm that works for you but how do we know we have improved in speed and how do we know we are always above the mark - a test will always tell us this

nbqa/__main__.py
neomatrix369: Enumerators allow lazy evaluation - to see if it helps or not the test benchmarks would help, I could be wrong and also if the number of items in the list are small, it may not be applicable but worth a try.

MarcoGorelli:
You're absolutely right, a performance suite would be ideal. So far I've just been using PyMC3's notebooks to check for performance, but I should really make that an automated suite that code checks are checked against

neomatrix369:
> You're absolutely right, a performance suite would be ideal. So far I've just been using PyMC3's notebooks to check for performance, but I should really make that an automated suite that code checks are checked against

It's very easy to done, just use my test as an example and then build on top of it.

I do both perf measurement and capture line profiling results and save it per test, my next goal is to parse this text and assert on the components but I haven;t got there yet

If you come up with something I'll be happy to help tidy it up (if necessary) but perf tests are important when you make speedups. Low-hanging fruits first and then those other ones which are suspects.

MarcoGorelli:
Cool, thanks!

The optimisation in this PR is definitely low-hanging fruit, but for future performance work where it's less clear, it'll be crucial to have something set up

nbqa/__main__.py
girip11: Given the size of `args` is going to be really small (since it holds the directories or notebooks to run nbqa) does it really matter whether its a tuple or list?  I am thinking in this case its just better to leave as it is since `args` is generated by list comprehension. Maybe using `args: Iterable[Path]`  would make for a better abstraction.

tests/test_return_code.py
```diff
@@ -15,6 +15,8 @@
 PASSED = 0
 # Failure indicates the QA tool reported some issues.
 FAILED = 1
+# pylint exits with 20 instead of 1
+PYLINT_FAILED = 20
```
girip11, line 19: Given we are starting to see different exit codes returned by these tools, I am suggesting to check `assert pylint_runner([DIRTY_NOTEBOOK]) != PASSED` for these scenarios. No need to add new exit codes in the future and so the tests would not break.

girip11 (changes requested):

nbqa/__main__.py
neomatrix369: I agree for the size it's not substantial to take advantage of my earlier suggestion.

nbqa/__main__.py
neomatrix369: So the benefit of enumerate is not just what you mention but the discussion won't be necessary as the size of the list is too small to take any advantage of these suggestions.

nbqa/__main__.py
neomatrix369: Good point

nbqa/__main__.py
MarcoGorelli: In PyMC3 this list can potentially have 80+ notebooks if you run nbqa via pre-commit, so I made the change to tuple anyway, thanks!

nbqa/__main__.py
```diff
@@ -270,11 +287,40 @@ def _get_arg(
...
+        Mapping between notebooks and Python files corresponding to converted notebooks.
+    project_root
+        Root of repository, where .git / .hg / .nbqa.ini file is.
+
+    Returns
+    -------
+    List[Path]
+        All notebooks or directories to run third-party tool against.
+    """
+    return tuple(
```
girip11, line 319: @MarcoGorelli : In order to use tuple, you have used `tuple(generator expression)` instead of earlier list comprehension. list comprehension performs better than tuple with generator expressions.

Ref: https://stackoverflow.com/questions/43808180/python-3-x-list-comprehension-vs-tuple-generator (refer accepted answer)

Also performance difference between tuple and lists is during creation only (also extra memory allocated for lists). While accessing both produce same performance.
Ref: https://stackoverflow.com/questions/68630/are-tuples-more-efficient-than-lists-in-python (refer accepted answer)

Based on your latest change, looks like you opted for less performant approach.

girip11 (changes requested):

nbqa/__main__.py
MarcoGorelli: :laughing: oops, thanks for catching me!

MarcoGorelli:
@girip11 have updated, any other comments?

girip11 (approved):

girip11:
> @girip11 have updated, any other comments?

Approved the PR

PR #373, opened 2020-10-24
ENH always pass 'treat comment as code' with modern isort

### LINKED ISSUE #163: use --treat-comment-as-code when running isort
see https://github.com/timothycrosley/isort/issues/1338#issuecomment-666918626

MarcoGorelli:
closes #163


nbqa/config.py
```diff
@@ -18,6 +23,31 @@ class _ConfigSections(NamedTuple):  # pylint: disable=R0903
...
+
+    In the case of isort, we use ``--treat-comment-as-code '# %%'``.
+
+    Returns
+    -------
+    Mapping
+        Default configs for each code quality tool.
+    """
+    default_configs = defaultdict(lambda: [])
+    if ISORT_MODULE is not None:
```
girip11, line 44: For the version check, we could use `importlib.metadata` with backporting lib for python 3.6 and 3.7.

```Python
from importlib.metadata import version

# maintain a dict of minimum versions supported for tools
min_ver_dict = {
"isort": "5.3.0"
}

# First thing we could after parsing the cli args is to check this in `main()`
command_version = version(command)
# this command version can be checked against in the min_ver_dict
# before even proceeding to run the entire procedure
```

## Reference

[importlib metadata usage](https://importlib-metadata.readthedocs.io/en/latest/using.html)

girip11: I thinking using `import_metadata` for version checks as well as collecting any other metadata is much better than we doing it ourselves.

Also we can use the same version to decide if the package is installed or not and return immediately. This is better for performance since we are not actually importing any module.

girip11 (changes requested):

nbqa/config.py
```diff
@@ -18,6 +23,31 @@ class _ConfigSections(NamedTuple):  # pylint: disable=R0903
 CONFIG_SECTIONS = _ConfigSections()
 
 
+try:
+    ISORT_MODULE: Optional[ModuleType] = import_module("isort")
+except ImportError:  # pragma: nocover (this case is tested in test_isort_works.py)
+    ISORT_MODULE = None
+
+
+def get_default_configs() -> Mapping[str, List[str]]:
```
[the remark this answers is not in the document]
MarcoGorelli: The option `--treat-comment-as-code` is only available for isort 5.3.0 onwards, so I think we would have to try importing `isort` to tell whether it's of a modern enough version or not

nbqa/config.py
MarcoGorelli: maybe this is better, actually...let's just "force" people to use isort 5.3.0+ and then we can always pass --treat-comment-as-code. Thanks for the suggestion, I'll make the changes!

girip11: For passing the config, please consider using a default config file as suggested in the comment above.

MarcoGorelli: Sure

> Along with importlib.resources in Python 3.7 and newer (backported as importlib_resources for older versions of Python), this can eliminate the need to use the older and less efficient pkg_resources package.

I guess this means we need to use `pkg_resources` if we want to support Python3.6?

MarcoGorelli: There isn't any way of easily comparing versions in `importlib_metadata`, is there? Or do we still need `from pkg_resources import parse_version`?

girip11: `importlib.metadata` is availalbe from python 3.8 onwards as part of stdlib. For python 3.6 and 3.7 we need to define `importlib_metadata; python_version < '3.8'` in setup.py `install_requires`.

## References
https://stackoverflow.com/questions/60517042/requirements-txt-conditional-on-python-version
https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers

girip11: > There isn't any way of easily comparing versions in importlib_metadata, is there? Or do we still need from pkg_resources import parse_version?

Simple string comparison would work right? `"5.6.4" > "5.3.0"` would work.

MarcoGorelli: @girip11 I've just been trying this, but I don't think it works if we have a local module we want to import.

E.g. I get the following failures:
```
FAILED tests/test_runtime_errors.py::test_unable_to_reconstruct_message - ModuleNotFoundError: Command `remove_comments` not found by nbqa.
FAILED tests/test_runtime_errors.py::test_unable_to_reconstruct_message_pythonpath - ModuleNotFoundError: Command `remove_comments` not found by nbqa.
FAILED tests/tools/test_doctest.py::test_doctest_works - ModuleNotFoundError: Command `doctest` not found by nbqa.
FAILED tests/tools/test_doctest.py::test_doctest_invalid_import - ModuleNotFoundError: Command `doctest` not found by nbqa.
```

IMO we should allow for running local scripts / modules on notebooks, although I can't see how to tell if they're importable with trying
```
    try:
        import_module(command)
```
Any suggestions?

girip11: In that case we have to distinguish between local script, standard module or python package. Shall we have option like `--script` to indicate it is a local script. This can allow for new modules that can be written by users too

MarcoGorelli: It should already work with custom modules though (e.g. see the test which runs `nbqa remove_comments`)

The overhead of importing modules doesn't seem very large
```
In [6]: %%timeit 
   ...: try: 
   ...:     import_module('foo') 
   ...: except: 
   ...:     pass 
   ...:                                                                                                                                                                                                   
58.8 µs ± 1.57 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
is it worth adding to the API to reduce the overhead of using `import_module`?

girip11: Adding these dependencies to nbqa at source code level which we claim to be tool_agnostic is not looking good when you have other ways right. Because then we would have more tool specific code creeping in to the source. So if we can go the generic way, even if its a hard way, its better to take it.

girip11: Also I think its good to separate module from packages right? What are your thoughts?

MarcoGorelli: > Because then we would have more tool specific code creeping in to the source

If I've understood correctly, I think we'd need _some_ tool-specific code anyway, because of
```
>>> from importlib.metadata import version
>>> version('doctest')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/importlib/metadata.py", line 531, in version
    return distribution(distribution_name).version
  File "/usr/lib/python3.8/importlib/metadata.py", line 504, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: doctest
```

> Also I think its good to separate module from packages right?

As in, to have a different way of running `nbqa <installed package>` from `nbqa <local module>`? If so, I'm not sure that'd be desirable, because when running Python scripts you can do both `python -m <installed package>` and `python -m <local module>`

girip11: > is it worth adding to the API to reduce the overhead of using import_module?

You are right. Hey how about using a fallback? 

```Python
# use importlib_metadata for checking version and module present or not.
# if we get result we can proceed straight way
# if we are not able to find version, try importlib.import_module
```

MarcoGorelli: Nice!! :+1: OK I'll do that, thanks!

girip11: Thanks @MarcoGorelli . Sometimes I get carried away like saying `handle module and package differently`. I completely agree with you that its not desirable

nbqa/config.py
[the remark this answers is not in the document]
s-weigand: If you want to include a data file, make sure to properly include it with `package_data` (been there done that, shipping a broken package because of a missing data file 😜 )
Here is a nice video explaining the datafile packaging differences
[](http://www.youtube.com/watch?v=OXmYKh0eTQ8 "")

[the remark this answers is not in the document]
MarcoGorelli: all praise our saviour Anthony Sottile

MarcoGorelli:
@nbQA-dev/nbqa-core I can't make sense of what's happening in the latest commit.

So, basically, if I run all tests like this, they all pass.

_but_ if I apply this change
```diff
$ git diff
diff --git a/tests/tools/test_isort_works.py b/tests/tools/test_isort_works.py
index f97fc23..60971e3 100644
--- a/tests/tools/test_isort_works.py
+++ b/tests/tools/test_isort_works.py
@@ -134,7 +134,8 @@ def test_isort_trailing_semicolon(tmp_notebook_with_trailing_semicolon: Path) ->
     path = os.path.abspath(
         os.path.join("tests", "data", "notebook_with_trailing_semicolon.ipynb")
     )
-    subprocess.run(["nbqa", "isort", path, "--nbqa-mutate"])
+    with pytest.raises(SystemExit):
+        main(["isort", path, "--nbqa-mutate"])
 
     with open(tmp_notebook_with_trailing_semicolon) as handle:
         after = handle.readlines()
```

and then run `tox -e py`, then I get:
```
_____________________________________________________________________________________________ test_pyupgrade _____________________________________________________________________________________________

tmp_notebook_for_testing = PosixPath('tests/data/notebook_for_testing.ipynb'), capsys = <_pytest.capture.CaptureFixture object at 0x7f1851f61070>

    def test_pyupgrade(tmp_notebook_for_testing: Path, capsys: "CaptureFixture") -> None:
        """
        Check pyupgrade works. Should only reformat code cells.
    
        Parameters
        ----------
        tmp_notebook_for_testing
            Temporary copy of :code:`tmp_notebook_for_testing.ipynb`.
        capsys
            Pytest fixture to capture stdout and stderr.
        """
        # check diff
        with open(tmp_notebook_for_testing) as handle:
            before = handle.readlines()
        path = os.path.abspath(os.path.join("tests", "data", "notebook_for_testing.ipynb"))
    
        Path("setup.cfg").write_text(
            dedent(
                """\
                [nbqa.mutate]
                pyupgrade = 1
    
                [nbqa.addopts]
                pyupgrade = '--py36-plus'
                """
            )
        )
        with pytest.raises(SystemExit):
            main(["pyupgrade", path])
        Path("setup.cfg").unlink()
        with open(tmp_notebook_for_testing) as handle:
            after = handle.readlines()
    
        result = "".join([i for i in diff if any([i.startswith("+ "), i.startswith("- ")])])
        expected = (
            "-    \"    return 'hello {}'.format(name)\\n\",\n"
            "+    \"    return f'hello {name}'\\n\",\n"
        )
>       assert result == expected
E       assert '' == ('-    "    return \'hello {}\'.format(name)\\n",\n'\n '+    "    return f\'hello {name}\'\\n",\n')
E         - -    "    return 'hello {}'.format(name)\n",
E         - +    "    return f'hello {name}'\n",

tests/tools/test_pyupgrade.py:57: AssertionError
------------------------------------------------------------------------------------------ Captured stderr call ------------------------------------------------------------------------------------------
usage: pyupgrade.py [-h] [--exit-zero-even-if-changed] [--keep-percent-format]
                    [--keep-mock] [--py3-plus] [--py36-plus] [--py37-plus]
                    [--py38-plus]
                    [filenames [filenames ...]]
pyupgrade.py: error: unrecognized arguments: --treat-comment-as-code # %%
======================================================================================== short test summary info =========================================================================================
FAILED tests/tools/test_mypy_works.py::test_mypy_works - assert [] == ['Found 3 errors in 3 files (checked 15 source files)',\n 'tests/data/notebook_for_testing.ipynb:cell_2:19: error: Argument 1 to ...
FAILED tests/tools/test_pylint_works.py::test_pylint_works - assert 0 == 4
FAILED tests/tools/test_pyupgrade.py::test_pyupgrade - assert '' == ('-    "    return \'hello {}\'.format(name)\\n",\n'\n '+    "    return f\'hello {name}\'\\n",\n')
====================================================================================== 3 failed, 46 passed in 8.57s ======================================================================================
ERROR: InvocationError for command /home/marco/nbQA-dev/.tox/py/bin/coverage run -m pytest tests -vv (exited with code 1)
________________________________________________________________________________________________ summary _________________________________________________________________________________________________
ERROR:   py: commands failed
```

See the `--treat-comment-as-code # %%` part? It's like it's being carried forward from one test to the next...have any of you come across anything like this before? 

I'm utterly confused

MarcoGorelli:
nevermind, figured it out :tada: will push soon

MarcoGorelli:
Essentially, we need to get `Configs.parse` to return a new `Configs` instance, otherwise `CLIArgs.parse_args(raw_args)` will be modified in place and this seems to cause problems during testing

requirements.txt
```diff
@@ -1 +1,2 @@
+importlib_metadata
```
girip11, line 1: It would be better to enforce it only for python < 3.8

nbqa/cmdline.py
```diff
@@ -66,8 +66,8 @@ def __str__(self) -> str:
 
         return " ".join(args)
 
-    @staticmethod
-    def parse_args(raw_args: Optional[List[str]]) -> "CLIArgs":
+    @classmethod
+    def parse_args(cls, raw_args: Optional[List[str]]) -> "CLIArgs":
```
girip11, line 70: Any reason for making it a class method?

nbqa/config_parser.py
```diff
@@ -113,16 +120,21 @@ def parse_config_from_file(cli_args: CLIArgs, project_root: Path) -> Optional[Co
...
-            break
+            # we found the config. skip other files
+            if config is not None:
+                break
+    else:
+        config = Configs()
+
+    for section in CONFIG_SECTIONS:
+        default = DEFAULT_CONFIG[section].get(cli_args.command, None)
+        if not getattr(config, section, None):
```
girip11, line 137: My understanding of this change is to always use `treat comment as code` option for isort. But in the above change if the user is passing some options to `isort` then `treat comment as code` option won't be set. I thought `treat comment as code` option would be appended to options passed by user. Basically `user_config.merge(default_config)`

girip11, line 137: `getattr(config, section, None)` will always return None because config instance does not contain `section` as it is.  All section members are prefixed with `_`.

nbqa/config_parser.py
```diff
@@ -96,6 +96,13 @@ def _parse_setupcfg_or_toxini_config(
     (".nbqa.ini", _parse_nbqa_ini_config),
 ]
 
+DEFAULT_CONFIG: Mapping[str, Mapping] = {
```
girip11, line 99: I will be simpler to store these configurations in `default.toml` and parse the config using existing `parse_from_pyproject_toml`. Then we can use `config.merge(default_config)`. This will be much simpler.

Its better not to have configuration as part of source code

girip11 (changes requested):

nbqa/config_parser.py
MarcoGorelli: oops, you're right, I'll just append it, thanks - I need a new test for this case then :)

nbqa/config_parser.py
MarcoGorelli: yes, you're right :+1: I'll do that in the next iteration

nbqa/config_parser.py
MarcoGorelli: @girip11 I was thinking about this again, and am _tempted_ to say that instead we should have some default configs for each tools which the user can then _overwrite_ (this'll be made explicit in the documentation - and if anyone will use pyproject.toml for configuring nbqa, it means they're already reading the docs :) ).

This way, for each tool, we can provide what we think are some sensible configs (similar to the ones we have in the examples from the readme), and the user can overwrite them if they wish. There may indeed be a reason someone doesn't want to use e.g. `--treat-comment-as-code` or any of the other default configs we will provide

girip11: If we are making the configuration as optional (that can be overridden by the user), then we really don't have to enforce isort version right? Since the PR title says `always pass treat comment as code`, I thought this change is adding a few configurations always on top of user provided configuration. Now we are at `Forcing some defaults vs Providing some sensible defaults`. If you are inclining towards providing sensible defaults rather than forcing those, then we should not raise exception when isort version is not above '5.3.0'. Rather we should do something like add these default configuration only when the tools version is more than so and so otherwise still run the tool. Could you elaborate more on this at issue #163 itself.

MarcoGorelli: :thinking: I should've thought this through before starting...I can't see why anyone would _not_ want `--treat-comment-as-code`, so perhaps we can treat this PR as "adding a few configurations always on top of user provided configuration". We can start with this `isort` one, and then perhaps extend it to others later (e.g. always ignore the 'trailing semicolon' error from flake8)

nbqa/cmdline.py
MarcoGorelli: Just cause it feels like this methods constructs an instance of the class, any reason not to? (I don't know much about OOP)

nbqa/config.py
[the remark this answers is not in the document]
MarcoGorelli: @s-weigand is that the video you meant to link? I couldn't figure out how to package a toml file from it, sorry :rofl:

nbqa/__main__.py
```diff
@@ -527,9 +530,19 @@ def _check_command_is_installed(command: str) -> None:
...
+        try:
+            import_module(command)
+        except ImportError as exc:
+            raise ModuleNotFoundError(_get_command_not_found_msg(command)) from exc
+    else:
+        if command in MIN_VERSIONS:
+            min_version = MIN_VERSIONS[command]
+            if parse_version(command_version) < parse_version(min_version):
+                raise ModuleNotFoundError(
+                    _get_command_not_found_msg(f"{command}>={min_version}")
```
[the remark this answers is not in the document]
MarcoGorelli: maybe...any suggestions?

nbqa/config_parser.py
MarcoGorelli: I haven't figured out how to package an extra toml file yet, but in the meantime, does the rest look alright to you?

girip11:
> I haven't figured out how to package an extra toml file yet, but in the meantime, does the rest look alright to you?

I have just left a few refactoring suggestions. Other than those minor ones, this change looks much better now.

girip11 (changes requested):

nbqa/config_parser.py
```diff
@@ -113,16 +113,16 @@ def parse_config_from_file(cli_args: CLIArgs, project_root: Path) -> Optional[Co
...
         if file_path.is_file() and config_handler is not None:
             config = config_handler(cli_args.command, file_path)
 
-        # we found the config. skip other files
-        if config is not None:
-            break
+            # we found the config. skip other files
+            if config is not None:
+                break
+    else:
```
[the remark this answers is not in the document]
MarcoGorelli: you're right, thanks!

nbqa/config.py
```diff
@@ -117,3 +129,21 @@ def parse_from_cli_args(cli_args: CLIArgs) -> "Configs":
         config.set_config(CONFIG_SECTIONS.MUTATE, cli_args.nbqa_mutate)
 
         return config
+
+    def merge_defaults(self, command: str) -> "Configs":
```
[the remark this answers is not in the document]
MarcoGorelli: That's neater, thanks! I'll apply that

MarcoGorelli:
Thanks for your careful review! I've updated - if it's OK, I'll leave figuring out how to package an extra toml file for a separate PR

tests/tools/test_isort_works.py
```diff
@@ -155,3 +156,55 @@ def test_isort_trailing_semicolon(tmp_notebook_with_trailing_semicolon: Path) ->
...
+    This test would fail if we didn't pass --treat-comment-as-code '# %%'.
+
+    Parameters
+    ----------
+    tmp_test_data
+        Temporary copy of test data.
+    """
+    notebook = tmp_test_data / "notebook_with_separated_imports_other.ipynb"
+
+    with open(notebook) as handle:
```
girip11, line 174: How about using Path.read_text()?
Or
How about using `os.path.getmtime` to see if the notebook was modified if the intent is to check if the notebook content got changed?

tests/tools/test_isort_works.py
```diff
@@ -155,3 +156,55 @@ def test_isort_trailing_semicolon(tmp_notebook_with_trailing_semicolon: Path) ->
...
+def test_old_isort(monkeypatch: "MonkeyPatch") -> None:
+    """
+    Check that using an old version of isort will raise an error.
+
+    Parameters
+    ----------
+    monkeypatch
+        Pytest fixture, we use it to override isort's version.
+    """
+    monkeypatch.setattr("nbqa.__main__.metadata.version", lambda _: "4.3.21")
```
girip11, line 200: Nice!

girip11 (approved):

girip11:
I have reviewed this PR. Good to merge.

> I'll leave figuring out how to package an extra toml file for a separate PR

Sure. Also shall we add a sensible defaults section for each official tool the nbqa supports to the documentation so that new users can refer to and readily use it?

MarcoGorelli:
Cool, thanks for your help here!

Issue #313, opened 2020-10-12
Add example / test for autoflake


MarcoGorelli:
I expect this one to work out-of-the box (for removing unused imports)

girip11:
Shall I take this up?

MarcoGorelli:
Sure, thanks! I've assigned you, but I think you have the permissions to assign yourself - is so, feel free to do so

girip11:
>  I think you have the permissions to assign yourself - is so, feel free to do so

Sure.

MarcoGorelli:
@girip11 I think autoflake would work unreliably if we have a notebook with e.g.

```python
# cell1
import pandas as pd

# cell2
%time pd.DataFrame()
```

where something is imported but is only used after an in-line magic.

An idea would be to raise an error if a notebook contains code after an inline magic and the user runs `nbqa autoflake`, advising them to instead run `nbqa flake8` and remove unused imports manually.

For `nbqa flake8` we can actually just replace `%time <code>` with `<code>` seeing as there's no need to reconstruct the notebook.

I can't think of a way to universally replace `%time <code>` with `<code>` in such a way that formatters will work without issue. All I can think of is to divide tools into linters (which don't modify notebooks) and formatters (which may modify notebooks), and for the former we replace  `%time <code>` with `<code>` and for the latter we replace  `%time <code>` with `type(hash) # <code> etc.` (as we currently do). Thoughts?

girip11:
I had come up with the following approach to fix issue #279. If I am able to fix that, the issue that you discussed above would be handled too.

## Approach

1.  Check if the current line is a line magic. If yes, check if the line magic contains a python code. This check is required because we could have something like `%time func()` or `%load_ext <ext_name>` etc.
2. In cases where the line magic does not contain a python code, we can use existing replacement statement which is `type(token)  # magic token`
3. In cases like line magics with python code (ex: `%time func()`), extract only the code and insert the code into the following template. This replacement code will be a multi line code. But since we use regex match in the latest code, we should be able to match across multiple lines.
```Python
# token or hash - both refer to the same thing
if int(token):
    <extracted_code> # token
```
Above code template would work with both linters and formatters. Formatters would format only `<extracted_code>` and our regex match would still work since we have fixed prefix(`if int(token):`) and fixed suffix(`# token`).

4. In case of linters nothing needs to be done further. 
[...]

PR #377, opened 2020-10-27
Handle python code present as part of line magics

girip11:
Takes care of issue #279 and also helps to support autoflake #313 .

- [ ] Test current changes on external notebooks

This PR is ready for review. But before merging I would like to run on some external notebooks.

### DIFF
--- a/nbqa/handle_magics.py
+++ b/nbqa/handle_magics.py
@@ -1,67 +1,11 @@
 """Detect ipython magics and provide python code replacements for those magics."""
+import ast
+import contextlib
 import re
 import secrets
 from abc import ABC
-from typing import Pattern, Tuple
-
-
-class MagicSubstitution:
-    """Store original ipython magic along with the replaced python code."""
-
-    replacement_line: str
-    _original_source: str
-    _substitution_pattern: Pattern[str]
-
-    def __init__(
-        self, replacement_line: str, original_source: str, pattern: Pattern[str]
-    ) -> None:
-        """
-        Initialize MagicSubstitution instance.
-
-        Parameters
-        ----------
-        replacement_line : str
-            Python code replacing the ipython magic
-        original_source : str
-            IPython magic present in the notebook cell
-        pattern : Pattern[str]
-            Regex pattern to restore the magic with python code
-        """
-        self.replacement_line = replacement_line
-        self._original_source = original_source
-        self._substitution_pattern = pattern
-
-    def restore_magic(self, source: str) -> str:
-        """
-        Replace the original magic in the source.
-
-        Parameters
-        ----------
-        source : str
-            Source code of the notebook cell
-
-        Returns
-        -------
-        str
-            Source of the notebook cell with original ipython magic restored.
-        """
-        return re.sub(self._substitution_pattern, self._original_source, source)
-
-    def indent_magic_replacement(self, spaces: str) -> str:
-        """
-        Indent the magic replacement code with the input leading spaces.
-
-        Parameters
-        ----------
-        spaces : str
-            Spaces to indent the replacement code.
-
-        Returns
-        -------
-        str
-            Indented replacement code
-        """
-        return f"{spaces}{self.replacement_line}"
+from ast import AST
+from typing import Optional, Pattern
 
 
 class MagicHandler(ABC):
@@ -75,47 +19,29 @@ class MagicHandler(ABC):
     # to double quotes or vice versa. `{token}` is used as hexadecimal number.
     _MAGIC_TEMPLATE: str = "type({token})  # {magic:10.10} {token}"
     _MAGIC_REGEX_TEMPLATE: str = r"type\s*\(\s*{token}\s*\).*{token}"
+    _token: str
+    _ipython_magic: str
 
-    def replace_magic(self, ipython_magic: str) -> MagicSubstitution:
+    def __init__(self, ipython_magic: str) -> None:
+        self._token: str = MagicHandler._get_unique_token()
+        self._ipython_magic = ipython_magic
+
+    def replace_magic(self) -> str:
         """
         Return python code to replace the ipython magic.
-
-        Parameters
-        ----------
-        ipython_magic : str
-            IPython magic statement present in the notebook cell
-
-        Returns
-        -------
-        MagicSubstitution
-            Python code to be substituted for the ipython magic
-        """
-        token: str = MagicHandler._get_unique_token()
-        replacement_line, pattern = self._get_magic_replacement(token, ipython_magic)
-        return MagicSubstitution(replacement_line, ipython_magic, pattern)
-
-    def _get_magic_replacement(
-        self, token: str, magic: str
-    ) -> Tuple[str, Pattern[str]]:
-        """
-        Return python code to be replace the input ipython magic.
-
-        Parameters
-        ----------
-        token : str
-            Token to uniquely identify the replacement python code
-        magic : str
-            IPython magic statement
 
         Returns
         -------
         str
-            Python code to replace the ipython magic
+            Python code to be substituted for the ipython magic
         """
-        return (
-            self._MAGIC_TEMPLATE.format(magic=magic, token=token),
-            MagicHandler._get_regex_pattern(self._MAGIC_REGEX_TEMPLATE, token),
+        return self._MAGIC_TEMPLATE.format(magic=self._ipython_magic, token=self._token)
+
+    def restore_magic(self, cell_source: str) -> str:
+        pattern = MagicHandler._get_regex_pattern(
+            self._MAGIC_REGEX_TEMPLATE, self._token
         )
+        return re.sub(pattern, self._ipython_magic, cell_source)
 
     @staticmethod
     def _get_unique_token() -> str:
@@ -177,14 +103,14 @@ class MagicHandler(ABC):
         """
         magic_handler: MagicHandler
         if ipython_magic[0] == "!":
-            magic_handler = ShellCommandHandler()
+            magic_handler = ShellCommandHandler(ipython_magic)
         elif ipython_magic[0] == "?" or ipython_magic[-1] == "?":
-            magic_handler = HelpMagicHandler()
+            magic_handler = HelpMagicHandler(ipython_magic)
         elif ipython_magic[0] == "%":
             if len(ipython_magic) > 1 and ipython_magic[1] == "%":
-                magic_handler = CellMagicHandler()
+                magic_handler = CellMagicHandler(ipython_magic)
             else:
-                magic_handler = LineMagicHandler()
+                magic_handler = LineMagicHandler(ipython_magic)
 
         return magic_handler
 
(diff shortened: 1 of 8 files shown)


girip11:
What is this `pre-commit.ci - pr` workflow ? This seems to be a new one. Did we enable this or onboard to this?

MarcoGorelli:
> What is this `pre-commit.ci - pr` workflow ? This seems to be a new one. Did we enable this or onboard to this?

am just trying out pre-commit-ci (alpha preview)

will review later, thanks for doing this!

girip11:
I just checked [pre-commit.ci](https://pre-commit.ci/). Interesting v1 features like **auto-commit back to pull request for changes** are in progress.

nbqa/handle_magics.py
```diff
@@ -210,4 +156,192 @@ class CellMagicHandler(MagicHandler):
...
+        cell_source : str
+            Source code of the notebook cell
+
+        Returns
+        -------
+        str
+            Cell source with ipython magic restored
+        """
+        ipython_magic = self._ipython_magic
+        with contextlib.suppress(AssertionError):
```
girip11, line 325: This method needs refactoring. Any better way to write without using if not None checks?

girip11:
After this PR, I can add tests to check if autoflake works to address 313.

MarcoGorelli:
I justed tested this on pymc3's notebooks, and all worked fine! :tada:

nbqa/handle_magics.py
```diff
@@ -210,4 +156,192 @@ class CellMagicHandler(MagicHandler):
...
+    statement, then flake8 or pylint would complain `unused import os`. Also tools
+    like autoflake would remove those statements from the notebook which is undesirable.
+
+    To handle this issue, we transform the above snippet to a temporary python code that
+    looks like the snippet below
+
+    ```Python
+    import os
+
+    if True
```
MarcoGorelli, line 186:
```suggestion
    if True:
```

nbqa/handle_magics.py
```diff
@@ -210,4 +156,192 @@ class CellMagicHandler(MagicHandler):
...
+    When ipython magic like `%time` that can contain python code, if we ignore the
+    code and replace the line magic with the template `type(hex_token)...`, we will
+    get into a scenario of unused imports.
+
+    For example, consider the following snippet.
+
+    ```Python
+    import os
+
+    if True
```
MarcoGorelli, line 172:
```suggestion
    if True:
```

tests/tools/test_black.py
```diff
@@ -210,3 +215,24 @@ def test_black_multiple_files(tmp_test_data: Path) -> None:
 
     diff = difflib.unified_diff(before, after)
     assert "".join(diff) != ""
+
+
+def test_successive_runs_using_black(tmpdir: "LocalPath") -> None:
+    """Check black returns 0 on the second run given a dirty notebook."""
```
MarcoGorelli, line 221: excellent test here!

MarcoGorelli (changes requested):
Awesome!!! Incredible work here

Rather than modifying `notebook_for_testing.ipynb`, would it be worth making a new test notebook and leaving that one intact?

girip11:
> Actually I'm not sure the output above works

@MarcoGorelli: Good catch! Yes its a known limitation. Line magics can take code that span multiple lines. That's one of the reasons I wanted to test out on external notebooks. Right now the simplest way(also the only way I can think of) would be to special case black as you mentioned. Later if we can think of better approach, we could refactor the current changes.

Currently I am suggesting something like `FORMATTERS = ['black', 'yapf', 'autopep8']` and skip special handling of line magics if the command is in the formatters list.

Linters flake8, pylint could also raise line length exceeded when the code looks like `%time <long line>`. But we can do anything there. I would like to test around that as well before merging this PR.

MarcoGorelli:
> Currently I am thinking on something like FORMATTERS = ['black', 'yapf', 'autopep8'] and skip special handling of line magics if the command is in the formatters list.

I think that's a good solution, I'm just wondering whether it would be better to make to default behaviour to _not_ use special handling of line magics and special-case pylint/flake8/etc. to use it. This is because false positives from an unknown linter are less bad than incorrect formatting from an unknown formatter

girip11:
> This is because false positives from an unknown linter are less bad than incorrect formatting from an unknown formatter

I found this [split line magic multiple lines](https://stackoverflow.com/questions/35021565/split-ipython-magic-shell-multiple-lines)
This works.

```Python
%time np.random(\
100\
)
```

@MarcoGorelli : Could you please confirm this once from your side?

image

MarcoGorelli:
Nice! Yes,
```
import numpy as np

%time np.random.randn(\
100\
)
```
works for me!

girip11:
> Rather than modifying notebook_for_testing.ipynb, would it be worth making a new test notebook and leaving that one intact?

I thought adding this test scenario to `notebook_for_testing.ipynb` made more sense because this file is extensively used across all tests and any regression can be caught better with this notebook. 

Also adding a separate test to test this change from the aspect of all the officially supported tools seemed more demanding.

nbqa/handle_magics.py
girip11: @MarcoGorelli: Do you see any drawback in using assert for testing `if something is not None` ?

MarcoGorelli:
ok sure, makes sense

nbqa/handle_magics.py
MarcoGorelli: Is the assert just there to check that there was a match? If so, Anthony Sottile has a video on some options: https://youtu.be/yH6L3UiK9WM

girip11:
@MarcoGorelli  @s-weigand : Need some help here with some manual testing. I have executed nbqa with these changes on all pymc3 notebooks. If you have any complex notebooks, could you please run nbqa with this change on those complex notebooks and see if you find the notebooks in good state after running nbqa?

girip11:
@MarcoGorelli : Could you run on pymc3 notebooks and run those to see if we get any issues while running the modified notebooks?

nbqa/handle_magics.py
girip11: > Is the assert just there to check that there was a match? 

Yes. Thanks I will check the video.

MarcoGorelli:
Incredible @girip11 !!! It seems to work as expected...I'm not aware of any other formatter/linter of ipynb that does this, you work here strikes me as truly exceptional.

In terms of aesthetics, maybe the intendation could be changed? e.g. if we have a notebook Untitled.ipynb with
```
import numpy as np

%time np.random.randn(1000)
```
and then run
```
nbqa black Untitled.ipynb --line-length=10
```
then we get
```
import numpy as np

%time np.random.randn(\
        1000\
    )
```

This is correct, and runs fine, but
```
import numpy as np

%time np.random.randn(\
          1000\
      )
```
also runs fine, and IMO looks a bit more like what people might expect to see

girip11:
> In terms of aesthetics, maybe the indentation could be changed?

Sure. Indented code would look much better. Will make those changes.

girip11:
> So, instead of 4 indents, I think the number of indents might be better as the length of the magic command + 1

@MarcoGorelli  I made those indentation changes. This PR is ready for review now.

girip11:
I am still testing with flake8, pylint. There is a bug with the line numbers mapping, fixable I think. @MarcoGorelli Lets keep this PR open for few days.

girip11:
> I am still testing with flake8, pylint. Issues with linter type tools are being addressed. There is a bug with the line numbers mapping due to the new way line magics with python code are handled (we expand a line magic to 2 lines using if block), fixable I think.

This is fixed now. I ran flake8 on 77 notebooks of pymc3 and verified the results manually on some notebooks. @nbQA-dev/nbqa-core Please proceed reviewing this change.

MarcoGorelli:
Yes, seems to be fixed now! PyMC3 only have one notebook which has an inline `%time`, so I've also done some manual testing with some crazy notebooks I've been trying to write, and...this seems to work exactly as intended!

Will keep reviewing more carefully as this is a pretty big change

MarcoGorelli:
Why do we need
```
with contextlib.suppress(AssertionError):
```
? It seems the code runs fine without it, and even if you suppress the assertion error, if `match_obj` is None then `replacement_code = match_obj.group()` will fail

MarcoGorelli:
Pending a decision on `with contextlib.suppress(AssertionError):`, I think this is good to go. From what I can tell, if there aren't any line magics, then the current behaviour is preserved. I think we can make a 0.4.0 release after this! Agree, or is there anything else we'd like to get in for that?

girip11:
> I think we can make a 0.4.0 release after this! Agree, or is there anything else we'd like to get in for that?

Agreed. No more changes need to go in except addressing review comments.

girip11:
> I think we can make a 0.4.0 release after this! Agree, or is there anything else we'd like to get in for that?

@MarcoGorelli: Since this is a big change, should we make something like a **beta release**(ex: `0.4b1`) instead of a stable release?

nbqa/handle_magics.py
girip11: Thanks @MarcoGorelli. The youtube video you shared was really helpful.

girip11:
Building docs failing because the link [LiuAlgoTrader .pre-commit-config.yaml](https://github.com/amor71/LiuAlgoTrader/blob/master/.pre-commit-config.yaml) no longer exists.

MarcoGorelli:
We already have

> Development Status :: 4 - Beta

in setup.cfg, so I'd be inclined to keep as is and then change to stable when we get to v1.0.0 if you agree

> Building docs failing because the link LiuAlgoTrader .pre-commit-config.yaml no longer exists.

:rofl: yeah, just noticed, though that library still has nbqa in their requirements. Perhaps we can just link to their requirements.txt instead, rather than removing them completely, but that's a separate issue, I'll ignore the docs-links failure here

girip11:
> in setup.cfg, so I'd be inclined to keep as is and then change to stable when we get to v1.0.0 if you agree

Agreed.

nbqa/handle_magics.py
MarcoGorelli: no worries - we don't need to suppress the assertionerror any more though, do we?

nbqa/handle_magics.py
```diff
@@ -210,4 +172,323 @@ class CellMagicHandler(MagicHandler):
...
+            if match_obj is None:  # pragma: nocover
+                raise AssertionError("unable to extract the replaced python code")
+
+            extracted_code = self._extract_code(match_obj.group())
+
+            # IPython magic is recreated from modified source so that formatters
+            # like black won't reformat the notebook every time they are run.
+            ipython_magic = f"""
+            {self._extract_line_magic()} {self._indent_extracted_code(extracted_code)}
+            """.strip()
```
MarcoGorelli, line 369: can the second line here be indented?

nbqa/handle_magics.py
```diff
@@ -210,4 +172,323 @@ class CellMagicHandler(MagicHandler):
...
+            Returns the line magic from the original source code.
+
+        Raises
+        ------
+        AssertionError
+            Raised if unable to extract the line magic from the source.
+        """
+        match_obj = re.match(r"%\w+", self._ipython_magic)
+        if match_obj is None:  # pragma: nocover
+            raise AssertionError("Unable to extract the line magic")
```
MarcoGorelli, line 443: we can do this in a separate PR, but IMO it would be nice to have tests which somehow get to these branches

girip11: Sure. I will try to cover those in a separate PR following this one.

nbqa/handle_magics.py
girip11: This looks better. I didn't know this will concatenate two strings.

nbqa/handle_magics.py
girip11: I wanted to catch the `AssertionError` and return a default value instead of throwing it to the user. What are  your thoughts?

MarcoGorelli: Part of me thinks that yes, we should throw the error back to the user, along with an encouragement to report a bug - what do you think?

s-weigand: IMHO it is nicer if a tool annoys me with spammy warnings than just exploding 😝

girip11: > Part of me thinks that yes, we should throw the error back to the user, along with an encouragement to report a bug - what do you think?

I am scared of this approach. Imagine if many such bugs are reported at the same time. I agree with your idea of encouraging the users to report such bugs. Nbqa has some projects already in pre-commit workflows. If I raise this error, it won't look good from our part right?

@s-weigand: I like your idea, but first is this what you are saying? Raise some warning from nbqa (write some message to stderr ) when we hit AssertionError?

So taking the middle ground -> I can raise some warnings from nbqa when AssertionError is raised and display well crafted warnings to the user with message to file bug report. But nbqa will handle this error gracefully and will proceed with further notebook processing.

MarcoGorelli: OK, thanks both for your input, let's go with a warning then :+1: and leave the return code intact

MarcoGorelli:
> @s-weigand: I like your idea, but let me confirm first if this what you are saying? Raise some warning from nbqa (write some message to stderr ) when we hit AssertionError codepath?
>
> I can raise some warnings from nbqa when AssertionError is raised and display well crafted warnings to the user with message to file bug report. But nbqa will handle this error gracefully and will proceed with further notebook processing. This would not impact github action workflows but users running it locally can see the warning and could report it.

@s-weigand just to confirm, is this what you had in mind? If so, I think we're ready for a 0.4.0 release once the warning's in, I think the enhancement this PR provides is really, really good (and, as far as I can tell, works as intended)

girip11:
I am working on raising warning when AssertionError code path is hit. I will send those changes out for PR by today.

nbqa/handle_magics.py
girip11: I updated the PR to raise a warning.

MarcoGorelli (approved):

nbqa/save_source.py
```diff
@@ -87,25 +89,87 @@ def _handle_magic_indentation(
...
+    """
+    if re.match(r"\s*%\w+", magic) is not None:
+        # Here we look for line magics spanning across multiple lines.
+        while magic.endswith(f"\\{NEWLINE}"):
+            try:
+                magic += next(cell_source)[1]
+            except StopIteration:  # pragma: nocover
+                # This scenario is a syntax error where a line magic
+                # ends with a backslash and does not have a next line.
+                break
```
s-weigand, line 166: Sorry for the (too) late reply.
What I had in mind was something in the direction.
warnings.warn(
    """Unable to extract magic from cell x.y, please check that the cell produces the desired output.

If it does, please file an issue at nbqa, with an example to reproduce the problem.
""")

That way we could inform users that something went wrong w/o breaking the CI if the code is fine and we have an internal error.

MarcoGorelli: it's OK, this could go in 0.4.1, thanks for your input!

PR #396, opened 2020-10-29
Add tests for docs

### LINKED ISSUE #345: [Question/Discussion] Deadlinks checker in documents across the repo/library
Just like I have this issue on https://github.com/neomatrix369/awesome-ai-ml-dl/issues/54, I have found a way to manually check for deadlinks in all of my markdown files in a repo.

I think there should be something that looks for such links in .rst and other document files as well.

When we find out, it would be good to add it to the `pre-commit` CI/CD GH action.

Can also be added to the local pre-commit.

In both cases it should be run when files that may contain hyperlinks are being committed

s-weigand:
- Fixed broken `setup.py`, for python < 3.8
- Adds Workflow that tests
  - If the docs build
  - If there are broken links
  - Uploads an artifact of the built docs
- Adds 2 new tox tests with the same functionality (`docs`, `docs-links`)
- Adds tox env to clean docs built artifacts cross platform (there will be errors if locally there are old API files, e.g. if a function gets renamed/removed)
- Makes doc creation more flexible (no more hardcoded module names, but auto discovery from the root package)
- Private methods/functions (starting with `_`) aren't part of the docs anymore

closes #345

### DIFF
--- a/docs/clean_build_artifacts.py
+++ b/docs/clean_build_artifacts.py
@@ -1,0 +1,14 @@
+"""Cross platform way to call 'rm -rf docs/_build/ docs/api/' """
+from pathlib import Path
+from shutil import rmtree
+
+
+def delete_artifacts():
+    """Delete API and _build directories"""
+    current_dir = Path(__file__).parent
+    rmtree(current_dir / "api", ignore_errors=True)
+    rmtree(current_dir / "_build", ignore_errors=True)
+
+
+if __name__ == "__main__":
+    delete_artifacts()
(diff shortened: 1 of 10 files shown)


docs/clean_build_artifacts.py
```diff
@@ -0,0 +1,14 @@
+"""Cross platfor way to call 'rm -rf docs/_build/ docs/api/' """
```
MarcoGorelli, line 1:
```suggestion
"""Cross platform way to call 'rm -rf docs/_build/ docs/api/' """
```

s-weigand: Ups... Fixed 😄

MarcoGorelli:
Nice! So with `tox -e docs` and `tox -e docs-links` we can test the docs

PR #403, opened 2020-10-30
Try using setup.cfg declarative metadata

MarcoGorelli:
Trying to imitate asottile https://github.com/asottile/pyupgrade

Seems I'd put loads of extra things (taken from some cookiecutter template) which might not have been needed?

@s-weigand as mentioned on gitter, I think you should be included as co-author - if you agree I'll do so

@girip11 if I've done this correctly, it lets us have the default configs as a separate file


setup.cfg
```diff
@@ -1,3 +1,48 @@
+[metadata]
+name = nbqa
+version = 0.3.6
+description = Run any standard Python code quality tool on a Jupyter Notebook
+long_description = file: README.md
+long_description_content_type = text/markdown
+url = https://github.com/nbQA-dev/nbQA
+author = Marco Gorelli, Girish Pasupathy
```
MarcoGorelli, line 8:
```suggestion
author = Marco Gorelli, Girish Pasupathy, Sebastian Weigand
```

girip11 (approved):

s-weigand:
## Problems with `bump2version` and `setup.cfg`
`bump2version` sadly doesn't support the `setup.cfg` only layout properly (see https://github.com/c4urself/bump2version/pull/83 and https://github.com/c4urself/bump2version/issues/87).
The other issue with `bump2version` would be that it most likely will have a fight with `setup-cfg-fmt` and `pre-commit-hooks/trailing-whitespace`, since it adds trailing whitespaces (see https://github.com/c4urself/bump2version/pull/161) and also rewrites the whole `setup.cfg` in config parser style (space indent -> tab indent).

I guess the only workaround to still use  `bump2version` would be to use `.bumpversion.cfg` file.

## Proposed `setup.cfg` changes
### Version
The version can be read automagically if we set `setuptools>=46.4.0` (released: May 17, 2020) as a requirement ([see the packaging guide](https://packaging.python.org/guides/single-sourcing-package-version/))
```ini
[metadata]
version = attr: package.__version__
```

## Project Urls
Having links to the docs and issues on PyPi is a nice to have ([should look something like this](https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#specifying-values))
```ini
[metadata]
project_urls =
        Documentation = https://nbQA.readthedocs.io/en/latest/
        Source = https://github.com/nbQA-dev/nbQA
        Tracker = https://github.com/nbQA-dev/nbQA/issues
```
It renders like this on PyPi
grafik


### Additional classifiers and keywords
I would also add the following classifiers
```ini
[metadata]
classifiers =
    Framework :: Jupyter
    Operating System :: OS Independent
    Environment :: Console
    Topic :: Software Development :: Libraries :: Python Modules
    Topic :: Software Development :: Quality Assurance
```
keywords

```ini
[metadata]
keywords = jupyter, notebook, format, lint, ...
```

setup.cfg
```diff
@@ -1,3 +1,48 @@
...
+
+[options.packages.find]
+exclude =
+    tests*
+
+[bdist_wheel]
+universal = True
+
 [bumpversion]
 current_version = 0.3.6
```
s-weigand, line 48: This can be removed one or the other way (see https://github.com/nbQA-dev/nbQA/pull/403#issuecomment-719933624)

s-weigand (changes requested):

nbqa/config_files/__init__.py
```diff
@@ -0,0 +1 @@
+"""Configuration files."""
```
s-weigand, line 1: We could just rename the package to `config` and also put the modules for the config here.

s-weigand (approved):

MarcoGorelli:
Thanks for your suggestions!

It's not too much effort to change version numbers, and there's usually some small changes that need to be done anyway, so I don't mind that not being automated just yet

PR #405, opened 2020-10-31
Encourage users to report bugs if we can't parse output

MarcoGorelli:
Also, have made some cosmetic changes to error reporting, see below for some examples:

file not found:

image

command not found:

image

Can't parse output:

image


girip11 (approved):

girip11:
Nice work!

PR #408, opened 2020-10-31
Add autoflake hook, and more

MarcoGorelli:
This was useful - I was wondering why pyupgrade was for Python>=3.6.1, but then running flake8 with this extension revealed
```
nbqa/config/config.py:15:5: TYP005 NamedTuple does not support defaults in 3.6.0
```


girip11 (approved):

s-weigand:
Just a little question. Where is the `meta` repo from? The hooks sound usefull so I want to check it out 😄

PR #410, opened 2020-10-31
Preserve tool-specific config files

### LINKED ISSUE #384: Preserve all supported config files for supported tools
In https://github.com/nbQA-dev/nbQA/blob/728d8b68440b46b1aa89292e3dfbb4ee42fb8916/nbqa/__main__.py#L20 we have
```python
CONFIG_FILES = ["setup.cfg", "tox.ini", "pyproject.toml"]
```

If anyone wants to use a config file which isn't listed there, then they need to manually provide via the `--nbqa-config` CLI argument, or by specifying it in the `pyproject.toml` file.

My suggestion would be to make `CONFIG_FILES ` a `defaultdict` which which returns tool-specific config files.

So, for example, I would expect to see
```console
>>> CONFIG_FILES["flake8"]
["setup.cfg", "tox.ini", ".flake8"]

>>> CONFIG_FILES["black"]
["pyproject.toml"]

>>> CONFIG_FILES["some_custom_tool"]
["setup.cfg", "tox.ini", "pyproject.toml"]
```

MarcoGorelli:
closes #384


nbqa/__main__.py
```diff
@@ -516,7 +544,9 @@ def _run_on_one_root_dir(
             for notebook in _get_all_notebooks(cli_args.root_dirs)
         }
 
-        _preserve_config_files(configs.nbqa_config, tmpdirname, project_root)
+        _preserve_config_files(
+            cli_args.command, configs.nbqa_config, tmpdirname, project_root
```
girip11, line 548: Would it be better to write it this way? 

```
config_files = (configs.nbqa_config and [configs.nbqa_config]) or CONFIG_FILES[cli_args.command]
_preserve_config_files(
      config_files, tmpdirname, project_root
```
Currently `_preserve_config_files` is looking like find configuration files and then preserve those. By passing the configuration files, `_preserve_config_files` this function is kind of simplified

girip11 (changes requested):

girip11 (approved):

nbqa/__main__.py
MarcoGorelli: that's better, thanks!

PR #411, opened 2020-11-01
Parse output from "unparseable error" from black

### LINKED ISSUE #399: Output from 'nbqa black' has line numbers referring to Python script, not Jupyter notebook
I was testing the command

```Python
nbqa black notebook.ipynb --nbqa-mutate
```

from [here](https://github.com/nbQA-dev/nbQA/issues/289#issuecomment-712199396) and while figuring out dependencies (black was missing) ended up having that command in a code cell in the active notebook I wanted to test the mutation on, with the path to the notebook (stated as ```correctpath``` below) originally mis-specified (stated as ```wrongpath``` below). 

When calling the command (from a terminal) on that notebook, with the path specified correctly, I got the error message

```Python
error: cannot format correctpath/notebook.ipynb: Cannot parse: 38:5: nbqa black wrongpath/notebook.ipynb --nbqa-mutate
Oh no! 💥 💔 💥
1 file failed to reformat.
```

I would expect nbQA to warn about the missing file, but not fail because of it.

MarcoGorelli:
closes #399

### DIFF
--- a/nbqa/output_parser.py
+++ b/nbqa/output_parser.py
@@ -32,6 +32,7 @@ def _get_pattern(
         Patterns and substitutions for reported output.
     """
     standard_substitution = partial(_line_to_cell, cell_mapping=cell_mapping)
+
     if command == "black":
         return [
             (
@@ -39,6 +40,7 @@ def _get_pattern(
                 standard_substitution,
             )
         ]
+
     if command == "doctest":
         return [
             (
@@ -47,6 +49,7 @@ def _get_pattern(
             ),
             (rf'(?<=^File "{re.escape(str(notebook))}",) line', ""),
         ]
+
     # This is the most common one and is used by flake, pylint, mypy, and more.
     return [(rf"(?<=^{re.escape(str(notebook))}:)\d+", standard_substitution)]
 
--- a/nbqa/output_parser.py
+++ b/nbqa/output_parser.py
@@ -3,7 +3,7 @@
 import re
 from functools import partial
 from pathlib import Path
-from typing import List, Mapping, Match, Tuple, Union
+from typing import Callable, List, Mapping, Match, Tuple, Union
 
 
 def _line_to_cell(match: Match[str], cell_mapping: Mapping[int, str]) -> str:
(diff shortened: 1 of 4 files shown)


MarcoGorelli:
failed CI job is likely due to codecov being down :cry:

nbqa/output_parser.py
```diff
@@ -0,0 +1,104 @@
...
+    str
+        Stderr with references to temporary Python file's lines replaced with references
+        to notebook's cells and lines.
+    """
+
+    def substitution(match: Match[str]) -> str:
+        """Replace Python line with corresponding Jupyter notebook cell."""
+        return str(cell_mapping[int(match.group())])
+
+    pattern = _get_pattern(notebook, command)
```
girip11, line 98: How about something like returning a list of patterns for a command as well as its replacements too and applying all those patterns sequentially one after the other on out and err?

```Python
def _get_pattern(command: str):
      # here we will have a dict for each command
      # replacement could be a callable or a string(or whatever re.sub works with)
      black = [("pattern1", replacement1), ("pattern2", replacement2)]
      doctest = [("pattern1", replacement1), ("pattern2", replacement2)]

# In map_python_line_to_nb_lines
patterns = _get_patterns(command)
for pattern in patterns:
      out = re.sub(pat[0], pat[1], out)
```

Something one the above lines?

girip11 (changes requested):

nbqa/output_parser.py
MarcoGorelli: That would be cleaner, thanks! I done something along those lines

nbqa/output_parser.py
```diff
@@ -0,0 +1,90 @@
...
+        Code quality tool.
+    cell_mapping
+        Mapping from Python file lines to Jupyter notebook cells.
+
+    Returns
+    -------
+    List
+        Patterns and substitutions for reported output.
+    """
+    standard_substitution = partial(_line_to_cell, cell_mapping=cell_mapping)
```
girip11, line 34: This change looks much cleaner now.

girip11 (approved):

PR #418, opened 2020-11-03
Clean up emoji usage

MarcoGorelli:


tests/tools/test_black.py
```diff
@@ -58,12 +59,13 @@ def test_black_works(tmp_notebook_for_testing: Path, capsys: "CaptureFixture") -
     # check out and err
     out, err = capsys.readouterr()
     expected_out = ""
-    expected_err = os.linesep.join(
-        [f"reformatted {path}", "All done!   ", "1 file reformatted."]
+    expected_err = (
+        fr"reformatted {re.escape(path)}\r?\n"
```
girip11, line 63: Can `\r?\n` be replaced with `\s{1,2}` ?

tests/tools/test_black.py
```diff
@@ -58,12 +59,13 @@ def test_black_works(tmp_notebook_for_testing: Path, capsys: "CaptureFixture") -
...
+        r"All done! (\\u2728|\u2728) (\\U0001f370|\U0001f370) (\\u2728|\u2728)\r?\n"
```
girip11, line 64: Thanks to this PR. I learnt this now. Looks like there is easier way to represent emojis in python. The representation you are using is unicode representation. Every emoji also has CLDR short name representation 

```python
# try this in repl
print("\U0001f370")

print("\N{shortcake}")
```

If we using this CLDR short name representation (this is human readable), the regex match becomes easier and the code is more readable too!

```Python
# declare this at module level, since this is used in multiple tests
SHORTCAKE_EMOJI= "\N{shortcake}"
SPARKLES = "\N{sparkles}"

# use your funny french strings here!
fr"All done! {SPARKLES} {SHORTCAKE_EMOJI} {SPARKLES}\s{1,2}"
```


---

## Reference

* [emojis full list](http://www.unicode.org/emoji/charts/full-emoji-list.html#1f370)
* [emojis in python](https://www.geeksforgeeks.org/python-program-to-print-emojis/)

girip11 (changes requested):

tests/tools/test_black.py
girip11: I tested locally and it worked.

```Python
    sparkles = "\N{sparkles}"
    shortcake = "\N{shortcake}"
    expected_err = (
        fr"reformatted {re.escape(path)}\s{{1,2}}"
        fr"All done! {sparkles} {shortcake} {sparkles}\s{{1,2}}"
        fr"{re.escape('1 file reformatted.')}\s{{1,2}}"
    )
    assert out == expected_out
    assert re.match(expected_err, err, re.S) is not None
```

But I have not tested with all python versions starting from 3.6. Hopefully it works in all python versions from 3.6.

MarcoGorelli: I didn't know about that, thanks!

Unfortunately, it doesn't seem to work on Windows, because the output is of the form
```
'No Path provided. Nothing to do \\U0001f634\r\n'
```
rather than
```
'No Path provided. Nothing to do \U0001f634\n'
```
(notice the extra backslash before `U0001f634`), which is why here I've put both possibilities

girip11: Oh ok. I may sound stupid here, but could out try fr"All done! \\?{sparkles} \\?{shortcake} \\?{sparkles}\s{{1,2}}" right?

s-weigand: Since this strongly depends on the font and shell used, IMHO there should be a flag to disable it (especially for Windows users).
I tested `python -c 'print("\U0001f634")'`:
WSL2+Terminal App (Font: Source Code Pro for Powerline): works 
WSL2 (Font: Lucida Console): fails
git-bash (Font: Consolas): fails
git-bash (Font: Source Code Pro for Powerline): fails
cmd (Font: Lucida Console): fails
powershell (Font: Lucida Console): fails

Especially for `cmd`, `powershell`, and vanilla WSL, it is nearly impossible, to get a nice font on windows due to the way windows filters the allowed fonts.

Maybe wrapp it in an `emoji` function and deactivate (`return ""`) it by default (at least on windows).

Or deactivate it by default in general (I don't know how the default fonts in different linux distros support emojis), and tell user to activate it if they want the cool emoji style (add screenshots).

MarcoGorelli: On Linux:
```
>>> import re
>>> sleep = "\N{sleeping face}"
>>> re.match(rf'{sleep}', '\U0001f634')
<re.Match object; span=(0, 1), match='😴'>
```
On Windows:
```
>>> import re
>>> sleep = "\N{sleeping face}"
>>> re.match(rf'\\?{sleep}', '\\U0001f634')
>>> re.match(rf'\?{sleep}', '\\U0001f634')
>>> re.match(rf'{sleep}', '\\U0001f634')
```

----

@s-weigand what happens if you just run `black` (with no extra arguments) in those environments? Is it OK to just return what the tool normally have returned, rather than adding an extra flag?

s-weigand: @MarcoGorelli using `black` there are the following issues:

WSL2+Terminal App (Font: Source Code Pro for Powerline): works ✔️ 
WSL2 (Font: Lucida Console): fails :x:
git-bash (Font: Consolas): fails :x:✔️ (🍰 can't be displayed)
git-bash (Font: Source Code Pro for Powerline): fails :x:✔️ (🍰 can't be displayed)
cmd (Font: Lucida Console): fails :x:
powershell (Font: Lucida Console): fails :x:

MarcoGorelli: OK that's fine, we can consider it an upstream limitation

s-weigand: Guess the title of the PR confused me 😅 
But we have the same issues with the emojis, nbqa itself uses (e.g. for error messages).

girip11: Since windows `cmd` does not have great support for unicode, I thought why not convert the `err` and `expectederr` to ascii with backslashes replaced. This way we can just compare all characters of the data.

```Python
    sparkles = "\N{sparkles}"
    shortcake = "\N{shortcake}"
    # replace \u with \\u for both expected_err and err
    expected_err = (
        (
            f"reformatted {path}{os.linesep}"
            f"All done! {sparkles} {shortcake} {sparkles}{os.linesep}"
            f"1 file reformatted.{os.linesep}"
        )
        .encode("ascii", "backslashreplace")
        .decode()
    )
    # This is required because linux supports emojis
    # so both should have \\ for comparison
    err = err.encode("ascii", "backslashreplace").decode()
    assert expected_err == err
```

I tested this on windows and linux.

girip11: I used black on windows cmd prompt. Black uses emojis but it cant be displayed in the cmd prompt. As @s-weigand  mentioned its the same case with nbqa too. 

@MarcoGorelli and @s-weigand : Should nbqa remove emojis or handle like @s-weigand mentioned(special case for windows)? I am fine with having emojis. Latest windows terminal anyways supports emojis. Also I  hope terminals like conemu or cmder might support emojis

MarcoGorelli: Nice!!!

@s-weigand that's a good point, maybe we should just remove them from our own messages then and only use them if the code quality tool does

s-weigand: Maybe just add a `--nbqa-emoji` flag 😝 
And tell users that this is deactivated for compat reasons, but hey look how cool it looks! 
In CI context it should work fine, which I guess is the most common usage besides `pre-commit`.

MarcoGorelli: If we only use them for error reporting, then IMO we might as well remove them...have done that in the latest commit

s-weigand: Also used here:
`� Mutation detected, will not reformat! Please use the `--nbqa-mutate` flag:`

MarcoGorelli: True...it pains to remove such a beautiful one, but it's probably for the better

MarcoGorelli: I've updated the demo, will push that change when we make the next release

girip11 (approved):

PR #423, opened 2020-11-05
Exclude .git, .venv, and other common non-source-code directories

MarcoGorelli:
Can be cleaned up later, but IMO it's prob. including this (or something like it) in the next release

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -48,7 +48,12 @@ MIN_VERSIONS = {"isort": "5.3.0"}
 VIRTUAL_ENVIRONMENTS_URL = (
     "https://realpython.com/python-virtual-environments-a-primer/"
 )
-EXCLUDES = r"/(\.ipynb_checkpoints|\.direnv|\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|venv|\.svn|_build|buck-out|build|dist)/"  # noqa: E501  # pylint: disable=C0301
+EXCLUDES = (
+    r"/("
+    r"\.direnv|\.eggs|\.git|\.hg|\.ipynb_checkpoints|\.mypy_cache|\.nox|\.svn|\.tox|\.venv|"
+    r"_build|buck-out|build|dist|venv"
+    r")/"
+)
 
 
 class UnsupportedPackageVersionError(Exception):
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -1,6 +1,7 @@
 """Run third-party tool (e.g. :code:`mypy`) against notebook or directory."""
 
 import os
+import re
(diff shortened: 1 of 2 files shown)


tests/test_get_notebooks.py
```diff
@@ -0,0 +1,27 @@
...
+    Parameters
+    ----------
+    monkeypatch
+        Pytest fixture, we use it to override ``PYTHONPATH``.
+
+    """
+    result = list(_get_notebooks("tests"))
+    assert Path("tests") / "data/clean_notebook.ipynb" in result
+
+    monkeypatch.setattr("nbqa.__main__.EXCLUDES", r"/(data)/")
```
girip11, line 25: This does not really test your EXCLUDES on code that will be shipped. Consider the scenario of having a bug in the `EXCLUDES` regex pattern. I would recommend to use `def test_get_notebooks(tmpdir: "LocalPath")`. Create `.git` dir inside this temp directory. Copy the notebook from `tests/data/clean_notebook.ipynb` to `tmpdir/.git`. You could do the same for many other directories like venv etc.

nbqa/__main__.py
```diff
@@ -47,6 +48,12 @@
 VIRTUAL_ENVIRONMENTS_URL = (
     "https://realpython.com/python-virtual-environments-a-primer/"
 )
+EXCLUDES = (
+    r"/("
+    r"\.direnv|\.eggs|\.git|\.hg|\.ipynb_checkpoints|\.mypy_cache|\.nox|\.svn|\.tox|\.venv|"
```
girip11, line 53: One way to simplify the regex pattern is to ignore the directories that start with `.`(hidden dirs). check if pattern like `r"/\.(\w|\s|-)+/"`

girip11 (changes requested):

tests/test_get_notebooks.py
MarcoGorelli: yes, that would be better, thanks!

nbqa/__main__.py
MarcoGorelli: true, but the fact that `black` doesn't do that makes me think we might run into something unexpected if we do that

girip11: I am wondering why would someone want to run a notebook inside a hidden directory?

girip11: If someone reports this as a bug/limitation, then we would know the scenario/usecase better right? That would be a better way to know about hidden dirs usage with notebooks.

s-weigand: IMHO using an `--exclude` flag/option and glob patterns with [`fnmatch`](https://docs.python.org/3/library/fnmatch.html) as flake8 does, might be more robust and common, than regex.

PR #430, opened 2020-11-07
Advise to overwrite additional_dependencies

MarcoGorelli:
The hook can still run with the latest version of whichever external tool it uses, but we should probably advise to pin the additional requirement

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,8 +118,11 @@ for details)
      rev: 0.4.0
      hooks:
        - id: nbqa-black
+         additional_dependencies: [black==20.8b1]
        - id: nbqa-pyupgrade
+         additional_dependencies: [pyupgrade==2.7.3]
        - id: nbqa-isort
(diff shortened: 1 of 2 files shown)


s-weigand:
Hmmm maybe mention that this leads to the best reproducibility while using the unpinned version uses the most recent version but might lead to different versions used.
To ensure the most recent versions you need to use `pre-commit clean` to delete the cached envs and the next time `pre-commit` runs it reinstalls them.

MarcoGorelli:
Thanks - I've added a note about how this leads to best reproducibility.

> To ensure the most recent versions you need to use pre-commit clean to delete the cached envs and the next time pre-commit runs it reinstalls them.

Not sure we need to document this if all the examples we show have pinned dependencies (after all, that's how people usually use tools, by copy-and-pasting from READMEs / docs)

MarcoGorelli:
merging, if anything else needs adding it can be done later

PR #432, opened 2020-11-07
Add --nbqa-files and --nbqa-exclude flags 

### LINKED ISSUE #429: Add --nbqa-files and --nbqa-exclude flags
currently nbqa moves files to a tmp dir before formatting - and so flake8's per-file-ignores cannot match files in the repository

MarcoGorelli:
closes #429 

----

still need to update the docs - the idea here is just to mirror the behaviour from pre-commit

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -536,7 +536,7 @@ def _run_on_one_root_dir(
         nb_to_py_mapping = {
             notebook: _temp_python_file_for_notebook(notebook, tmpdirname, project_root)
             for notebook in _get_all_notebooks(
-                cli_args.root_dirs, cli_args.nbqa_files, cli_args.nbqa_exclude
+                cli_args.root_dirs, configs.nbqa_files, configs.nbqa_exclude
             )
         }
 
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -91,9 +91,42 @@ def _get_notebooks(root_dir: str) -> Iterator[Path]:
     )
 
 
-def _get_all_notebooks(root_dirs: List[str]) -> Iterator[Path]:
+def _filter_by_include_exclude(
+    notebooks: Iterator[Path],
+    include: Optional[str],
+    exclude: Optional[str],
+) -> Iterator[Path]:
     """
-    Get generator with all notebooks passed in via the command-line.
+    Include files which match include, exclude those matching exclude.
+
+    notebooks
+        Notebooks (not directories) to run code quality tool on.
+    include:
+        Global file include pattern.
+    exclude:
+        Global file exclude pattern.
+
+    Returns
+    -------
+    Iterator
+        Notebooks matching include and not matching exclude.
+    """
+    include = include or ""
+    exclude = exclude or "^$"
+    include_re, exclude_re = re.compile(include), re.compile(exclude)
+    return (
+        notebook
+        for notebook in notebooks
+        if include_re.search(str(notebook.as_posix()))
+        if not exclude_re.search(str(notebook.as_posix()))
+    )
+
+
+def _get_all_notebooks(
+    root_dirs: List[str], files: Optional[str], exclude: Optional[str]
+) -> Iterator[Path]:
+    """
+    Get generator with all notebooks passed in via the command-line, applying exclusions.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


nbqa/__main__.py
```diff
@@ -105,7 +138,9 @@ def _get_all_notebooks(root_dirs: List[str]) -> Iterator[Path]:
     Iterator
         All Jupyter Notebooks found in all passed directories/notebooks.
     """
-    return (j for i in root_dirs for j in _get_notebooks(i))
+    return _filter_by_include_exclude(
+        (j for i in root_dirs for j in _get_notebooks(i)), files, exclude
```
girip11, line 142: Use `nbqa-files` patterns in `_get_notebooks` function as part of `Path.rglob()` and use `--nbqa-exclude` in `_filter_by_exclude`.

girip11 (changes requested):

MarcoGorelli:
Hi @nbQA-dev/nbqa-core 

The idea here is simply to mirror `files` and `excludes` from pre-commit https://pre-commit.com/#top_level-files

Code quality tools aren't consistent in how they implement include/exclude:
- [pydocstyle](http://www.pydocstyle.org/en/stable/usage.html?highlight=match#cli-usage) uses regular expressions;
- [flake8](https://flake8.pycqa.org/en/latest/user/options.html?highlight=--exclude#cmdoption-flake8-exclude) uses glob patterns;
- [black](https://black.readthedocs.io/en/stable/installation_and_usage.html) takes a regular expression, but _only_ when running on a directory (IMO this is the most confusing of them all)

So the idea here is to do _exactly_ what pre-commit does with `files` and `exclude`. My thinking here is that:
- pre-commit is arguably the best way to run linters/formatters, so it's good to mirror what it does (plus we copy their code);
- if we just mirror pre-commit, we can link to their docs and our behaviour won't feel like something new;
- the approach taken by pre-commit (regex on posix-like file names) is very flexible.

nbqa/__main__.py
MarcoGorelli: If we use regular expressions then I don't think this'll work - plus, like this, we can copy [filter_by_exclude](https://github.com/pre-commit/pre-commit/blob/e836e554891beba0e8e51092116559004772bf68/pre_commit/commands/run.py#L61) from pre-commit, which we know works

girip11:
Thanks @MarcoGorelli  for the clarification. Now I understood why you have used regex. It completely makes sense to be consistent with pre-commit.

girip11 (approved):

nbqa/__main__.py
```diff
@@ -91,9 +91,42 @@ def _get_notebooks(root_dir: str) -> Iterator[Path]:
...
+        Global file exclude pattern.
+
+    Returns
+    -------
+    Iterator
+        Notebooks matching include and not matching exclude.
+    """
+    include = include or ""
+    exclude = exclude or "^$"
+    include_re, exclude_re = re.compile(include), re.compile(exclude)
```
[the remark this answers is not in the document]
s-weigand: Was about to write pretty much the same comment yday, but when I checked our supported tolls I saw that only `flake8`, `mypy` and `isort` use glob patterns (`isort` even has an extra flag for it). While `pre-commit`, `pylint` and `black` use RegEx.
So we already have a mixed way to exclude files in the QA ecosystem, and since `pre-commit` is the recommendet way to use `nbqa`, people will have to get more familiar with RegEx anyway.
Thus I did write a comment sugesting glob patterns instead. 😜

nbqa/__main__.py
s-weigand: Btw. don't forget to give props to the `pre-commit` code 😉 
E.g. inspired by <permalink_to_pre-commit_code>

nbqa/__main__.py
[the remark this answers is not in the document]
s-weigand: Just saw @MarcoGorelli wrote pretty much the same train of thought below 🤣

nbqa/__main__.py
MarcoGorelli: you're right, I gotta do that!

MarcoGorelli:
Cool, thanks - have added some notes to the configuration page, if you have further comments on them please let me know, else I'll merge in a day or two

girip11 (approved):

nbqa/__main__.py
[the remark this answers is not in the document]
girip11: > Was about to write pretty much the same comment yday, but when I checked our supported tolls I saw that only flake8, mypy and isort use glob patterns (isort even has an extra flag for it). While pre-commit, pylint and black use RegEx.

I just looked at flake8 since it was mentioned in the issue #429 and hence raised this question. But going with pre-commit way is the right thing to do given how well nbqa is made for pre-commit usage.

girip11:
> Cool, thanks - have added some notes to the configuration page, if you have further comments on them please let me know, else I'll merge in a day or two

:+1:  Changes look good to me.

PR #436, opened 2020-11-09
Use universal_newlines=True

MarcoGorelli:
If this works it could simplify things.

Also, I think the current behaviour isn't quite right on Windows, as it adds extra newlines if you do e.g.
```
nbqa flake8 tests > out.txt
```
even though it looks fine when printing to the console.

Like this (with universal_newlines=True) then it looks correct on Windows both on the console, and when piping to a text file. Behaviour on Linux is unchanged.

This is great as it means we can simplify quite a few tests!

----

Just to clarify what I mean:

If I run
```
black tests --check 2> out.txt
```
then in `out.txt` I see:
```
would reformat C:\Users\m.gorelli.CORP\nbQA-dev\tests\test_find_root.py
would reformat C:\Users\m.gorelli.CORP\nbQA-dev\tests\test_ipython_magics.py
Oh no! \U0001f4a5 \U0001f494 \U0001f4a5
2 files would be reformatted, 26 files would be left unchanged.
```

However, if I write a script `subprocessrun.py` with
```python
import sys
import subprocess

output = subprocess.run(['black', 'tests', '--check'], stderr=subprocess.PIPE)
sys.stderr.write(output.stderr.decode())
```
and then run
```
python subprocessrun.py 2> out.txt
```
then in `out.txt` I see
```
would reformat C:\Users\m.gorelli.CORP\nbQA-dev\tests\test_find_root.py

would reformat C:\Users\m.gorelli.CORP\nbQA-dev\tests\test_ipython_magics.py

Oh no! \U0001f4a5 \U0001f494 \U0001f4a5

2 files would be reformatted, 26 files would be left unchanged.

```

If I change the Python script to
```
import sys
import subprocess

output = subprocess.run(['black', 'tests', '--check'], stderr=subprocess.PIPE, universal_newlines=True)
sys.stderr.write(output.stderr)
```
then `out.txt` looks correct:
```
would reformat C:\Users\m.gorelli.CORP\nbQA-dev\tests\test_find_root.py
would reformat C:\Users\m.gorelli.CORP\nbQA-dev\tests\test_ipython_magics.py
Oh no! \U0001f4a5 \U0001f494 \U0001f4a5
2 files would be reformatted, 26 files would be left unchanged.
```

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -398,14 +398,15 @@ def _run_command(
         stdout=subprocess.PIPE,
         cwd=tmpdirname,
         env=env,
+        text=True,
     )
 
     mutated = [_get_mtimes(i) for i in args] != before
 
     output_code = output.returncode
 
-    out = output.stdout.decode()
-    err = output.stderr.decode()
+    out = output.stdout
+    err = output.stderr
 
     return out, err, output_code, mutated
 
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -398,7 +398,7 @@ def _run_command(
         stdout=subprocess.PIPE,
         cwd=tmpdirname,
         env=env,
-        text=True,
+        universal_newlines=True,  # from Python3.7 this can be replaced with `text`
     )
 
     mutated = [_get_mtimes(i) for i in args] != before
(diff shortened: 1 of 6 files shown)


s-weigand:
Does it also look fine on less smart windows shells like `cmd` and `powershell`?

MarcoGorelli:
yup!

MarcoGorelli:
merging then, as this means I can implement --nbqa-diff

PR #442, opened 2020-11-11
Autoflake support

girip11:


girip11:
When I run the tests locally using the command `pytest tests/test_runtime_errors.py`, `test_runtime_errors:test_missing_root_dir` is failing. This is expected `__main__.py: error: the following arguments are required: root_dirs`, but the actual message is `pytest: error: the following arguments are required: root_dir`. But if I run the same tests using `python -m pytest tests/test_runtime_errors.py` the tests are passing. Also when I run `nbqa black`, the message displayed is `nbqa: error: the following arguments are required: root_dirs`.  @nbQA-dev/nbqa-core Any idea why this is happening? How are the tests passing at your end? Tests are passing in CI as well.

MarcoGorelli:
ah, that failure happens because I usually just run `tox -e py`, so `__main__.py: error:` is what appears. changing it to a regex so we can simply call `pytest` seems like a good idea

MarcoGorelli:
Anyway, nice PR! Do you also want to add a pre-commit hook for this?

girip11:
> Anyway, nice PR! Do you also want to add a pre-commit hook for this?

Yes, I am planning to raise a different PR with autoflake pre-commit hook and docs updated.

MarcoGorelli:
awesome - thanks!

Issue #445, opened 2020-11-11
iPython syntax protection


henryiii:
It seems some valid iPython syntax breaks this:

```
error: cannot format /private05-pybind/05-pybind.ipynb: Cannot parse: 8:7: pyinc = !python3-config --cflags
error: cannot format /private05-pybind/05b-pybind.ipynb: Cannot parse: 11:6: mkdir pyminuit2
```

I can make the mkdir magic explicit, %mkdir, but the `x = !...` line doesn't have an obvious fix.

MarcoGorelli:
Thanks @henryiii for the report!

I'm not sure there's a simple fix here, we _may_ have to treat `x = !...` as a known limitation... @girip11 any thoughts?

henryiii:
I would assume that `!...` is protected already, so `item = !...` should also be protected in a simliar way? I know that `!...` by itself doesn't break things.

henryiii:
Basically this is not quite true: https://github.com/nbQA-dev/nbQA/blob/036332e96865994060159ac30d16dd5cdf29cc16/nbqa/handle_magics.py#L128 - `item = !` is also supported in IPython.

MarcoGorelli:
You're right, we should fix this really...we'll keep it on the roadmap, thank you so much for having brought it to our attention!

In the meantime, perhaps this works for you as a workaround:
```
>>> import subprocess
>>> output = subprocess.run(['python3-config', '--cflags'], stdout=subprocess.PIPE, text=True)
>>> item = output.stdout
>>> item
'-I/usr/include/python3.6m -I/usr/include/python3.6m  -Wno-unused-result -Wsign-compare -g -fdebug-prefix-map=/build/python3.6-0aiVHW/python3.6-3.6.9=. -specs=/usr/share/dpkg/no-pie-compile.specs -fstack-protector -Wformat -Werror=format-security  -DNDEBUG -g -fwrapv -O3 -Wall\n'
```

s-weigand:
Another workaround would be to use `_`, which is set to the last returned value.
```python
In [1]: "dummy return"
Out[1]: 'dummy return'

In [2]: foo = _

In [3]: foo
Out[3]: 'dummy return'
```

As for the general problem, we might be able to solve it using `nbconvert.filters.ipython2python` on whole cells, this also partially solves the 'indented line magic' problem.

A little example of how it would look like in this case.
```python
In [1]: from nbconvert.filters import ipython2python                                     
                                                                                         
[...]

PR #450, opened 2020-11-12
Preserve ipython magics.

girip11:
Addresses bug #445

Changes to be committed:
	modified:   nbqa/handle_magics.py
	modified:   setup.cfg

### DIFF
--- a/tests/test_ipython_magics.py
+++ b/tests/test_ipython_magics.py
@@ -1,7 +1,8 @@
 """Check user can check for other magics."""
-import json
+import difflib
 from pathlib import Path
 from shutil import copyfile
+from textwrap import dedent
 from typing import TYPE_CHECKING, Callable, List, Optional, Tuple
 
 import pytest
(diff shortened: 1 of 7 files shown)


girip11:
closes #445

girip11:
Tested on pymc3 notebooks as well as the notebooks present in the [henryiii/python-compiled-minicourse](https://github.com/henryiii/python-compiled-minicourse).

girip11:
This PR does not address #455. But changes made in this PR makes it easier to fix #455.

MarcoGorelli:
> This PR does not address #455. But changes made in this PR makes it easier to fix #455.

Awesome! So this PR changes e.g.
```
flake8_version = !pip list 2>&1 | grep flake8
```
into
````
type(0xCDB9ABB2)  # flake8_ver 0xCDB9ABB2
```
, which won't throw any errors but which may still throw a false-positive (as per #455).

Looks good - do you have an idea already about how to transform the code to take care of #455?

tests/test_ipython_magics.py
```diff
@@ -123,69 +124,116 @@ def test_ignore_cells(
...
+def _validate_magics_with_black(before: List[str], after: List[str]) -> bool:
+    """
+    Validate the state of the notebook before and after running nbqa with black.
+
+    Parameters
+    ----------
+    before
+        Notebook contents before running nbqa with autoflake
+    after
+        Notebook contents after running nbqa with autoflake
```
MarcoGorelli, line 136: I'm a bit confused here - the summary says "before and after running nbqa with black", but the param descriptions refer to autoflake

MarcoGorelli (approved):

tests/test_ipython_magics.py
girip11: Good catch. copy pasted. Forgot to change. I will change it to black.

MarcoGorelli: cool, feel free to merge after that

girip11:
> - do you have an idea already about how to transform the code to take care of #455?

Yeah I have a rough idea to handle this issue. I will comment on the issue about the approach that I have.

tests/test_ipython_magics.py
girip11: Sure

PR #452, opened 2020-11-12
Add --nbqa-diff

### LINKED ISSUE #352: add --nbqa-diff flag
Display diff _without_ applying it

MarcoGorelli:
closes #352 

Example output:
---
image

----

@Daniel-Mietchen you'd said you'd find this feature useful for teaching, so am just tagging you here in case you have any input

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -598,6 +598,10 @@ def _run_on_one_root_dir(
                         )
                     ) from exc
 
+        if configs.nbqa_diff:
+            sys.stdout.write('To apply changes (if any), use `--nbqa-mutate` instead of `--nbqa-diff`\n')
+            return 0
+
         sys.stdout.write(out)
         sys.stderr.write(err)
 
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -337,10 +337,8 @@ def _get_arg(
     'tmpdir/my_notebook.py'
     """
     if Path(root_dir).is_dir():
-        arg = Path(tmpdirname) / Path(root_dir).resolve().relative_to(project_root)
-    else:
-        arg = nb_to_py_mapping[Path(root_dir)]
-    return arg
+        return Path(tmpdirname) / Path(root_dir).resolve().relative_to(project_root)
+    return nb_to_py_mapping[Path(root_dir)]
 
 
 def _get_all_args(
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -642,9 +642,10 @@ def _run_on_one_root_dir(
                     ) from exc
 
         if configs.nbqa_diff:
-            sys.stdout.write(
-                "To apply changes (if any), use `--nbqa-mutate` instead of `--nbqa-diff`\n"
-            )
+            if mutated:
+                sys.stdout.write(
+                    "To apply changes use `--nbqa-mutate` instead of `--nbqa-diff`\n"
+                )
             return 0
 
         sys.stdout.write(out)
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -563,7 +563,7 @@ def _run_on_one_root_dir(
                 sys.stderr.write(BASE_ERROR_MESSAGE.format(msg))
 
             if mutated:
-                if not configs.nbqa_mutate:
+                if not configs.nbqa_mutate and not configs.nbqa_diff:
                     # pylint: disable=C0301
                     msg = dedent(
                         f"""\
@@ -576,8 +576,10 @@ def _run_on_one_root_dir(
                     raise SystemExit(msg)
 
                 try:
-                    replace_source.main(
-                        temp_python_file, notebook, nb_info_mapping[notebook]
+                    replace_source.mutate(
+                        temp_python_file,
+                        notebook,
+                        nb_info_mapping[notebook],
                     )
                 except Exception as exc:
                     raise RuntimeError(
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -10,7 +10,7 @@ from collections import defaultdict
 from importlib import import_module
 from pathlib import Path
 from textwrap import dedent
-from typing import DefaultDict, Dict, Iterator, List, Optional, Set, Tuple
+from typing import DefaultDict, Dict, Iterator, List, Optional, Set, Tuple, NamedTuple
 
 from pkg_resources import parse_version
 
@@ -54,6 +54,15 @@ EXCLUDES = (
     r"_build|buck-out|build|dist|venv"
     r")/"
 )
+class Choice(NamedTuple):
+    diff: bool
+    mutate: bool
+
+
+REPLACE_FUNCTION = {
+    Choice(diff=True, mutate=False): replace_source.diff,
+    Choice(diff=False, mutate=True): replace_source.mutate,
+}
 
 
 class UnsupportedPackageVersionError(Exception):
@@ -575,8 +584,9 @@ def _run_on_one_root_dir(
                     # pylint: enable=C0301
                     raise SystemExit(msg)
 
+                replace_fn = REPLACE_FUNCTION[Choice(diff=configs.nbqa_diff, mutate=configs.nbqa_mutate)]
                 try:
-                    replace_source.mutate(
+                    replace_fn(
                         temp_python_file,
                         notebook,
                         nb_info_mapping[notebook],
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -10,7 +10,7 @@ from collections import defaultdict
 from importlib import import_module
 from pathlib import Path
 from textwrap import dedent
-from typing import DefaultDict, Dict, Iterator, List, NamedTuple, Optional, Set, Tuple
+from typing import DefaultDict, Dict, Iterator, List, Optional, Set, Tuple
 
 from pkg_resources import parse_version
 
@@ -56,16 +56,9 @@ EXCLUDES = (
 )
 
 
-class Choice(NamedTuple):
-    """Replace function choice."""
-
-    diff: bool
-    mutate: bool
-
-
 REPLACE_FUNCTION = {
-    Choice(diff=True, mutate=False): replace_source.diff,
-    Choice(diff=False, mutate=True): replace_source.mutate,
+    True: replace_source.diff,
+    False: replace_source.mutate,
 }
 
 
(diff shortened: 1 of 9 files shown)


nbqa/replace_source.py
```diff
@@ -87,24 +169,116 @@ def main(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) ->
...
+
+    pycells = _get_pycells(python_file)
+
+    code_cell_number = 0
+    for cell in notebook_json["cells"]:
+        if cell["cell_type"] == "code":
+            code_cell_number += 1
+            new_source = _get_new_source(code_cell_number, notebook_info, pycells)
+            if new_source is not None:
+                add_newline_to_last_line(cell["source"])
```
girip11, line 276: It would be better to check `if not cell["source"] and not new_source: `, so that the `unified_diff` function call can be simplified as 
```Python
cell_diff = unified_diff(
                    cell["source"].append("\n"),
                    new_source.append("\n"),
                    fromfile=str(notebook),
                    tofile=str(notebook),
                )
```

and `add_newline_to_last_line` function can be removed

nbqa/replace_source.py
```diff
@@ -87,24 +169,116 @@ def main(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) ->
...
+    cell
+        Current cell being processed.
+    """
+    if cell:
+        cell[-1] += "\n"
+
+
+def diff(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) -> None:
+    """
+    Replace :code:`source` code cells of original notebook.
```
girip11, line 255: Correct the function description `Replace :code:`source` code cells of original notebook.`

nbqa/replace_source.py
```diff
@@ -87,24 +169,116 @@ def main(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) ->
...
+            elif line.startswith("+"):
+                sys.stdout.write(print_green(line))
+            elif line.startswith("-"):
+                sys.stdout.write(print_red(line))
+            else:
+                sys.stdout.write(line)
+        sys.stdout.write("\n")
+
+
+def add_newline_to_last_line(cell: List[str]) -> None:
```
girip11, line 240: This function can be removed

nbqa/replace_source.py
```diff
@@ -87,24 +169,116 @@ def main(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) ->
...
-            if code_cell_number not in notebook_info.code_cells_to_ignore:
-                source = _restore_semicolon(
-                    next(pycells), code_cell_number, notebook_info.trailing_semicolons
-                )
-                cell["source"] = _reinstate_magics(
-                    source,
-                    notebook_info.temporary_lines.get(code_cell_number, []),
-                )
-
+            new_source = _get_new_source(code_cell_number, notebook_info, pycells)
```
girip11, line 179: (redundant comment) This could also be refactored once the changes suggested for `diff` function gets implemented.

* generator to give only the code cells
*  `_get_new_source(code_cell_number, notebook_info, pycells)` becomes ` _get_new_source(code_cell_number, notebook_info, next(pycells))` when the for loop becomes
```Python
 for cell_no, cell in enumerate(_notebook_code_cells, start=1):
    if cell_no not in notebook_info.code_cells_to_ignore:
        #entire logic follows here
```

nbqa/replace_source.py
```diff
@@ -87,24 +169,116 @@ def main(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) ->
...
+    Parameters
+    ----------
+    code_cell_number
+        Current cell number
+    cell_diff
+        Diff between original and new versions of cell.
+    """
+    peek, cell_diff = _peek(cell_diff)
+    if peek is not None:
+        header = f"Cell {code_cell_number}"
```
girip11, line 225: You could maintain a list and add all the lines you want to that list and finally use `sys.stdout.writelines(your_list)`. I am thinking this would be faster compared to calling `sys.stdout.write` for every line.

Also with this you could remove the `_peek` function.

```Python
line_changes = []
for line in cell_diff:
	if line.startswith("+++") or line.startswith("---"):
                line_changes.append(print_green(line))
	elif line.startswith("+"):
 		line_changes.append(print_green(line))
         elif line.startswith("-"):
 		line_changes.append(print_green(line))
         else:
 		line_changes.append(print_green(line))

if line_changes:
	header  = f"Cell {code_cell_number}"
	headers = [
       f"{BOLD}{header}{RESET}\n"),
       f"{'-'*len(header)}\n"
	]
	sys.stdout.writelines(headers + line_changes)
```

nbqa/replace_source.py
```diff
@@ -4,16 +4,46 @@
...
 from nbqa.save_source import CODE_SEPARATOR
 
 if TYPE_CHECKING:
     from pathlib import Path
 
+RED = "\033[31m"
+GREEN = "\033[32m"
+
+
+def _peek(iterable: Iterator[str]) -> Tuple[Optional[str], Iterator[str]]:
```
girip11, line 25: better to rename as `iterable` as `cell_diff`

girip11 (changes requested):

girip11:
It would be a very useful feature. I like the overall implementation :+1: , though I have recommended some refactorings

nbqa/__main__.py
```diff
@@ -601,21 +605,27 @@ def _run_on_one_root_dir(
                 sys.stderr.write(BASE_ERROR_MESSAGE.format(msg))
 
             if mutated:
-                if not configs.nbqa_mutate:
+                if not configs.nbqa_mutate and not configs.nbqa_diff:
```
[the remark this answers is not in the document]
MarcoGorelli: agreed, it's getting too big...I'd be more inclined to do it as a follow-up

nbqa/replace_source.py
```diff
@@ -72,7 +102,59 @@ def _reinstate_magics(
...
+    python_file
+        Temporary Python file notebook was converted to.
+
+    Returns
+    -------
+    Iterator
+        Parsed cells.
+    """
+    pyfile = python_file.read_text()
+    pycells: Iterator[str] = iter(pyfile[len(CODE_SEPARATOR) :].split(CODE_SEPARATOR))
```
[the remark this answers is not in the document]
MarcoGorelli: much better, thanks!

nbqa/replace_source.py
MarcoGorelli: are you sure this would work? because `cell["source"].append("\n")` returns `None`, e.g.
```
>>> [1,2,3].append('\n') is None
True
```

nbqa/__main__.py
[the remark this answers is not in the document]
girip11: I am also in favor of taking the refactoring in another PR. I just wanted to point out about the refactoring in this PR.

nbqa/replace_source.py
girip11: You are right. I changed `cell["source"] + ["\n"]` to use append but failed to move it out of the `unified_diff` function call.  You could replace `add_newline_to_last_line` with `cell["source"].append("\n")`. That was the intention.

s-weigand: Adding one more line above with
```python
cell["source"].append("\n"), new_source.append("\n")
```
And just using `cell["source"]` and `new_source` in `unified_diff` should work though.

MarcoGorelli: @s-weigand are you sure? I still don't think it's the same - currently, I'm adding a newline to the last element of a list, rather than adding an extra element which is a newline.

i.e. if we start with
```
['foo\n', 'bar']
```
I want to end up with
```
['foo\n', 'bar\n']
```
rather than
```
['foo\n', 'bar', '\n']
```

nbqa/replace_source.py
```diff
@@ -87,24 +169,116 @@ def main(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) ->
...
+    Parameters
+    ----------
+    python_file
+        Temporary Python file notebook was converted to.
+    notebook
+        Jupyter Notebook third-party tool is run against (unmodified).
+    notebook_info
+        Information about notebook cells used for processing
+    """
+    notebook_json = json.loads(notebook.read_text())
```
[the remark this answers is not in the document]
MarcoGorelli: @girip11 you're right, this could be greatly simplified! Thanks :+1: !

nbqa/replace_source.py
girip11: Now I understood. This `['foo\n', 'bar\n']` is how the source code looks inside `cell["source"]`. So what you did was the right way I feel, but still you don't need a method for it right?

MarcoGorelli: Probably not :smile: method removed

girip11:
@MarcoGorelli: Please leave a reply here when all the review comments are addressed.

nbqa/replace_source.py
s-weigand: @MarcoGorelli My comment was a comment to https://github.com/nbQA-dev/nbQA/pull/452#discussion_r523024054 . the gihub page didn't autorefresh 😝

s-weigand:
Btw is there a specific reason for using `sys.stdout.write` instead of `print`?
If not we could use [`rich`](https://github.com/willmcgugan/rich) to color and format the output.

MarcoGorelli:
> @MarcoGorelli: Please leave a reply here when all the review comments are addressed.

@girip11 sure, I think I've addressed everything

> Btw is there a specific reason for using sys.stdout.write instead of print?

No reason :smile: TBH My only concern is that then we introduce extra dependencies, which slows down install time and risks introducing "dependency hell"...unless `rich` makes output look noticeably better, I'd be hesitant to introduce it

girip11:
Approving the PR

s-weigand:
With `rich` we could get support for `cmd`, `powershell` and `gitbash` with don't properly support ANSI coloring.

MarcoGorelli:
> With `rich` we could get support for `cmd`, `powershell` and `gitbash` with don't properly support ANSI coloring.

Nice! OK, I'll look into it and will update (or, if it's something you've worked with before, do you want to do it in a follow-up PR?)

MarcoGorelli:
I'll merge this for now, using `rich` can be a follow-up and although I agree it would be nice-to-have, it's not a blocker for a release

s-weigand:
Didn't use `rich` yet, it is just one of my 500 stared repos as a 'that looks nice and might come in handy some day' 😆

PR #460, opened 2020-11-13
Validate config

### LINKED ISSUE #451: Raise error if --nbqa-config specifies non-existent file

MarcoGorelli:
closes #451 

still need to write some little docstrings for the tests

### DIFF
--- a/nbqa/config/config.py
+++ b/nbqa/config/config.py
@@ -25,10 +25,9 @@ CONFIG_FILES: DefaultDict[str, List[str]] = defaultdict(
     lambda: ["setup.cfg", "tox.ini", "pyproject.toml"]
 )
 CONFIG_FILES["black"] = ["pyproject.toml"]
-CONFIG_FILES["flake8"] = ["setup.cfg", "tox.ini", ".flake8"]
-CONFIG_FILES["pyupgrade"] = []
-CONFIG_FILES["mypy"] = ["mypy.ini", ".mypy.ini", "setup.cfg"]
+CONFIG_FILES["check-ast"] = []
 CONFIG_FILES["doctest"] = []
+CONFIG_FILES["flake8"] = ["setup.cfg", "tox.ini", ".flake8"]
 CONFIG_FILES["isort"] = [
     ".isort.cfg",
     "pyproject.toml",
@@ -36,7 +35,9 @@ CONFIG_FILES["isort"] = [
     "tox.ini",
     ".editorconfig",
 ]
+CONFIG_FILES["mypy"] = ["mypy.ini", ".mypy.ini", "setup.cfg"]
 CONFIG_FILES["pylint"] = ["pylintrc", ".pylintrc", "pyproject.toml", "setup.cfg"]
+CONFIG_FILES["pyupgrade"] = []
 
 
 class _ConfigSections(NamedTuple):  # pylint: disable=R0903
--- a/nbqa/config/config.py
+++ b/nbqa/config/config.py
@@ -24,6 +24,7 @@ from nbqa.cmdline import CLIArgs
 CONFIG_FILES: DefaultDict[str, List[str]] = defaultdict(
     lambda: ["setup.cfg", "tox.ini", "pyproject.toml"]
 )
+CONFIG_FILES["autoflake"] = []
 CONFIG_FILES["black"] = ["pyproject.toml"]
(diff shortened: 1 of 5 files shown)


nbqa/config/config.py
```diff
@@ -208,3 +246,28 @@ def validate(self) -> None:
                     """
                 )
             )
+        if self.nbqa_config and not Path(self.nbqa_config).exists():
+            raise FileNotFoundError(f"{self.nbqa_config} not found.")
+        if (
+            self.nbqa_config
```
girip11, line 252: This condition can be simplified to 

```Python
if command in CONFIG_FILES.keys() and self.nbqa_config not in CONFIG_FILES[command]
```

instead of checking self.nbqa_config, we can check once at the top. Also this validate method could be split in to smaller methods

```Python
def validate(self, command):
      _validate_nbqa_args()
      _validate_config_file_exists()
      _validate_config_file_for_command()
```

nbqa/config/config.py
```diff
@@ -1,14 +1,44 @@
...
+CONFIG_FILES["isort"] = [
+    ".isort.cfg",
+    "pyproject.toml",
+    "setup.cfg",
+    "tox.ini",
+    ".editorconfig",
+]
+CONFIG_FILES["mypy"] = ["mypy.ini", ".mypy.ini", "setup.cfg"]
+CONFIG_FILES["pylint"] = ["pylintrc", ".pylintrc", "pyproject.toml", "setup.cfg"]
+CONFIG_FILES["pyupgrade"] = []
```
girip11, line 40: Could you add it for autoflake too if there's any that it supports?

girip11 (changes requested):

nbqa/config/config.py
MarcoGorelli:
```
if command in CONFIG_FILES.keys() 
```
problem is, this is a default dict, so just accessing an element adds it to the keys.

e.g.:
```
>>> from collections import defaultdict
>>> d = defaultdict(list)
>>> d.keys()
dict_keys([])
>>> d[9]
[]
>>> d.keys()
dict_keys([9])
```

MarcoGorelli:
On a second thought, I'm not even sure if we want this...raising an error if the file's not found might be OK, but raising if the config file passed isn't one supported by the tool might not be.

E.g. `flake8` allows you to pass a custom config file:
```
  --config CONFIG       Path to the config file that will be the authoritative config source. This will cause Flake8 to ignore all other configuration files.
```

I'm just trying to weigh the benefit of increased flexibility against the possibility of confusion... @nbQA-dev/nbqa-core any thoughts here would be appreciated :smile:

girip11:
> On a second thought, I'm not even sure if we want this...raising an error if the file's not found might be OK, but raising if the config file passed isn't one supported by the tool might not be.
> 
> E.g. `flake8` allows you to pass a custom config file:
> ```
>   --config CONFIG       Path to the config file that will be the authoritative config source. This will cause Flake8 to ignore all other configuration files.
> ```
> 
> I'm just trying to weigh the benefit of increased flexibility against the possibility of confusion... @nbQA-dev/nbqa-core any thoughts here would be appreciated :smile: 

Great point! I feel this condition to check if we have passed the correct configuration file should be taken care by that tool and not nbqa. I am strongly in favor of removing this check

nbqa/config/config.py
girip11: But `command in config_files. keys()` would not add it to the dictionary, because we are not checking directly on the dict rather on the result returned by the keys method

girip11: I am in favor of removing this check.

pyproject.toml
```diff
@@ -28,4 +28,5 @@ disable = [
   "E0401", # Unable to import '_pytest.capture' (this is just for typing anyway)
   "E1136", # Value 'Optional' is unsubscriptable (bug? https://github.com/PyCQA/pylint/issues/3882)
   "E0239", # Inheriting 'NamedTuple', which is not a class. (bug? https://github.com/PyCQA/pylint/issues/2570)
+  "E1102", # CONFIG_FILES.default_factory is not callable (not-callable) (it is)
```
girip11, line 31: This rule can be removed now.

girip11 (approved):
I left a minor comment to remove the `E1102` rule from pyproject.toml. Otherwise I am fine with the changes made. Approving this PR.

PR #469, opened 2020-11-18
Fix linter warnings when shell magic assignment statements are present

### LINKED ISSUE #455: Fix linter warnings raised when ! magic output is assigned to a variable
Given below code snippet

```Python
flake8_version = !pip list 2>&1 | grep flake8

if flake8_version:
    print(flake8_version)
```

Mypy, flake8 and pylint raise `undefined name 'flake8_version'`

girip11:
closes #455

### DIFF
--- a/nbqa/handle_magics.py
+++ b/nbqa/handle_magics.py
@@ -282,14 +282,28 @@ class ShellCommandHandler(MagicHandler):
 
     def replace_magic(self) -> str:
         """
-        Return python code to be replace the input ipython magic.
+        Return python code that replaces the input ipython magic.
+
+        With shell magics, one can assign the output of the shell command to a python
+        variable. Consider the following snippet as the source of a notebook cell.
+
+        .. code:: python
+
+            flake8_version = !pip list 2>&1 | grep flake8
+            if flake8_version:
+                print(flake8_version)
+
+        If we use the usual template ``type(hex_code)`` to replace this shell magic,
+        linters would raise warnings like "Name 'flake8_version' is not defined". For
+        such shell magics, the python code template used is
+        ``var_name = ascii(shell_command) # hex_token``, so that the linters would be
+        happy.
 
         Returns
         -------
         str
             Python code to replace the ipython magic
         """
-        # check if the output of the shell command should be assigned to a variable
         assign_cmd_output_pattern = MagicHandler._MAGIC_PREFIXES[
             IPythonMagicType.SHELL
         ][1]
--- a/nbqa/handle_magics.py
+++ b/nbqa/handle_magics.py
@@ -287,7 +287,7 @@ class ShellCommandHandler(MagicHandler):
         With shell magics, one can assign the output of the shell command to a python
         variable. Consider the following snippet as the source of a notebook cell.
 
-        .. code:: python
+        .. code:: ipython3
 
             flake8_version = !pip list 2>&1 | grep flake8
             if flake8_version:
--- a/nbqa/handle_magics.py
+++ b/nbqa/handle_magics.py
@@ -252,7 +252,6 @@ class ShellCommandHandler(MagicHandler):
     _cmd = ""
     _var = ""
 
-    # r"\s*\w+\s*=\s*get_ipython(\s*).getoutput(\s*'(.*)'\s*)\s*"
     _EXTRACT_COMMAND_PATTERN: str = r"getoutput\('(.*)'\)"
     _EXTRACT_VARNAME_PATTERN: str = r"(\w+)\s*="
     _REPLACE_MAGIC_TEMPLATE: str = r'{var} = ascii(""" {command} """)  # {token}'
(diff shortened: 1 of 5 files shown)


girip11:
I would like to add more documentation to the `ShellCommandHandler` class.

nbqa/handle_magics.py
```diff
@@ -148,18 +184,9 @@ def is_ipython_magic(source: str) -> bool:
...
-            # contains any `get_ipython` function call.
-            # If the statement is a valid ipython magic, then ipython2python
-            # will transform using another `get_ipython()` function call.
-            src_count = source.count("get_ipython()")
-            warnings.simplefilter("ignore", DeprecationWarning)
-            # see https://github.com/nbQA-dev/nbQA/issues/459
-            return ipython2python(source).count("get_ipython()") == (1 + src_count)
+        return MagicHandler._ipython2python(source).count(
+            "get_ipython()"
+        ) > source.count("get_ipython()")
```
MarcoGorelli, line 189: barring refactoring, the only change here is
```diff
- == 1 + src_count
+ > src_count
```

(note to self :smile: )

nbqa/handle_magics.py
```diff
@@ -221,6 +248,89 @@ class HelpMagicHandler(MagicHandler):
...
     """Handle ipython magic containing !."""
 
+    _assign_shell_output = False
+    _cmd = ""
+    _var = ""
+
+    _EXTRACT_COMMAND_PATTERN: str = r"getoutput\('(.*)'\)"
+    _EXTRACT_VARNAME_PATTERN: str = r"(\w+)\s*="
+    _REPLACE_MAGIC_TEMPLATE: str = r'{var} = ascii(""" {command} """)  # {token}'
+    _RESTORE_MAGIC_REGEX_TEMPLATE: str = r"{var}\s*=\s*ascii.*{token}"
```
MarcoGorelli, line 258: wow, nice idea!

MarcoGorelli:
Awesome work! Let me know when it's ready for review - further documenting/commenting can always come in a separate PR if you want

girip11:
>  Let me know when it's ready for review - further documenting/commenting can always come in a separate PR if you want

@MarcoGorelli : I have updated the issue #455 with the approaches I had thought of. The problem with those approaches is that when we have a long shell command that itself takes up 70-100 as its line length, then black will keep reformatting the notebook. That is the reason I didn't make the PR ready for review. Do you think we should go ahead with this PR?

MarcoGorelli:
> The problem with those approaches is that when we have a long shell command that itself takes up 70-100 as its line length, then black will keep reformatting the notebook.

Hey - I couldn't reproduce this, I did
```bash
(nbqa-dev) mgorelli@m-gorelli02:~/nbqa-dev$ python -m nbqa black tests/data/notebook_with_indented_magics.ipynb --nbqa-mutate
reformatted tests/data/notebook_with_indented_magics.ipynb
All done! ✨ 🍰 ✨
1 file reformatted.
```
and the relevant cell in the notebook has become
```python
flake8_version = ascii(
    """ pip list 2>&1                                                  | grep                                                            flake8 """
)  # 0x877481CE
```

If I then run ` python -m nbqa black tests/data/notebook_with_indented_magics.ipynb --nbqa-mutate` again, then it doesn't reformat, even though the string is now really long.

So it looks like something's gone wrong during substituion - however, maybe instead of making the replacement
```
r'{var} = ascii(""" {command} """)  # {token}'
```
we could do something like
```
r'{var} = ascii(\n    """ {command} """\n)  # {token}'
```
? I haven't tried this yet, but I think that `black` is OK with long strings if they're on their own line?

girip11:
> `r'{var} = ascii(\n    """ {command} """\n)  # {token}'`

I have to check if black does not reformat the code when the entire line can fit within the configured line length.

girip11:
I have confirmed that black reformats the code every time when I used `r'{var} = ascii(\n """ {command} """\n) # {token}'` and the actual line length was less than the max line length configured for black.

I need to do some more thinking it seems. Marking this PR as draft.

MarcoGorelli:
> > `r'{var} = ascii(\n    """ {command} """\n)  # {token}'`
> 
> I have to check if black does not reformat the code when the entire line can fit within the configured line length.

Oh yeah, you're right...this is a tricky one to deal with, I think we _may_ have to treat it as a known limitation. The person who reported the bug with shell magic assignment was only using `black` and `pyupgrade` anyway IIRC - in fact, autoformatting seems to be a much more common use-case for nbqa than linting.

One suggestion would be that we _could_ make a 0.5.0 release, with support for shell magic assignment (though with this known limitation for linter warnings)

girip11:
@MarcoGorelli  We could go ahead and release next version of nbqa. I will keep making progress on this PR. As you pointed out, formatting tools are majorly used with nbqa. So we could treat the issue with the linter as a known limitation.

girip11:
closing this pull request and raising a new one when ready

PR #478, opened 2020-11-22
Remove reqs dev

MarcoGorelli:


s-weigand:
IMHO it't is nice to have a `requirements-dev.txt` so you can setup your dev environment outside of `tox` as well.
Without you would need to manually install the dev dependencies if you don't use `tox` to create your venv.
E.g. I use `conda` to create my dev envs and manage them with [`enboard`](https://pypi.org/project/enboard/).

MarcoGorelli:
OK, that's a pretty good point - let's keep it as is then, thanks!

PR #484, opened 2020-11-23
Handle ipython help magics when used in assignment expression

girip11:


nbqa/handle_magics.py
```diff
@@ -144,6 +144,15 @@ def is_ipython_magic(source: str) -> bool:
         """
         Return True if the source contains ipython magic.
 
+        A statement like ``%%timeit get_ipython().run_line_magic("magic", "input")``
+        already contains ``get_ipython()`` function call. To handle such cases, we need
+        to check if the original source itself contains any ``get_ipython()`` function
+        call before concluding the statement is an ipython magic.
+
+        Statement like ``some_result = str.split??`` gets transformed into two
+        statements of python code when using ``ipython2python``. Thus we need to check
```
MarcoGorelli, line 153: perhaps now replace `ipython2python` with the class we're using?

MarcoGorelli:
cool, thanks for this (and for #485)! Reckon we should do a 0.5.1 release with them?

girip11:
>  Reckon we should do a 0.5.1 release with them?

Yes we could do a 0.5.1 release.

PR #490, opened 2020-11-29
fix: UnicodeDecodeError

sbrugman:
nbQA helps us to keep notebooks to the same standards as the rest of the code. If you're serious about your code standards, you should keep them consistent across both notebooks and python scripts. Great addition to the ecosystem, thanks!

Version nbQA 0.5.1 throws an UnicodeDecodeError exception on my Windows machine. The proposed changes in this PR resolve it for my configuration, however I could imagine it's desirable to configure/automatically detect the charset for more general usage.

Find the updated configuration of how we use nbQA in pandas-profiling here: https://github.com/pandas-profiling/pandas-profiling/pull/628

### DIFF
--- a/nbqa/replace_source.py
+++ b/nbqa/replace_source.py
@@ -157,7 +157,7 @@ def mutate(python_file: "Path", notebook: "Path", notebook_info: NotebookInfo) -
     notebook_info
         Information about notebook cells used for processing
     """
-    notebook_json = json.loads(notebook.read_text())
+    notebook_json = json.loads(notebook.read_text(encoding="utf-8"))
 
     pycells = _get_pycells(python_file)
     for code_cell_number, cell in enumerate(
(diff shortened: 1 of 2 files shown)


s-weigand (approved):
LGTM Thanks for the PR

Don't know how many times I had this problem on windows, where python used the codepage from the console encoding and assumed that files use the same encoding.

```python
import this

...
Explicit is better than implicit.
...
```

MarcoGorelli:
Thanks @sbrugman , both for your kind words (mind if we include them as a testimonial?) and for this fix

MarcoGorelli:
@all-contributors add @sbrugman for bugs

sbrugman:
> mind if we include them as a testimonial?

Not at all, feel free @MarcoGorelli

sbrugman:
Thanks for the quick merge

MarcoGorelli:
cool, thanks! Have made a patch release (0.5.2) with this fix

PR #497, opened 2020-12-04
Fix Strange behaviour when applying nbqa black to commented ipython magic 

### LINKED ISSUE #495: Strange behaviour when applying nbqa black to commented ipython magic
## Behavior
Running `nbqa black` on [this file](https://gist.github.com/john-sandall/b128acba562cc020fe089635426ee9fd) shifts cells' contents into the next cell down. Seems to be related to the presence of `# %%capture`.

## Steps to Reproduce
1. Run `nbqa black '1 - Minimal nbqa black breaking example 1 - goes nuts.ipynb'` on the file linked above.
2. Notice that the contents of cell 5 are now in cell 6, with the commented ipython magic cell `# %%capture` now uncommented with the `%%` removed. The contents of cell 6 are now in cell 7, but the output of cell 7 hasn't changed.
3. We noticed that this only happens if a) there's a cell in the notebook that that black needs to reformat (hence the cell at the top), and b) a commented out `# %%capture` is present.
4. Removing either the top cell (so there's no black formatting applied) or uncommenting the capture magic so it's just `%%capture` runs as expected with no funky cell switcheroo.

## Environment
- macOS 10.14.6 (18G6032)
- Python 3.8.6
- nbqa 0.5.2
- ipython 7.19.0, black 20.8b1

## Mitigations
We were using `%%capture` to inhibit output from matplotlib. We now call `plt.close()` within our plotting code instead.

MarcoGorelli:
closes #495

### DIFF
--- a/nbqa/save_source.py
+++ b/nbqa/save_source.py
@@ -18,7 +18,7 @@ from nbqa.notebook_info import NotebookInfo
 if TYPE_CHECKING:
     from pathlib import Path
 
-CODE_SEPARATOR = "# %%"
+CODE_SEPARATOR = "# %%\n"
 MAGIC = [
     "%%!",
     "%%bash",
@@ -232,7 +232,7 @@ def _parse_cell(
         Parsed cell.
     """
     substituted_magics: List[MagicHandler] = []
-    parsed_cell = f"{CODE_SEPARATOR}{NEWLINE}"
+    parsed_cell = CODE_SEPARATOR
 
     for parsed_line in _replace_magics(source, substituted_magics):
         parsed_cell += parsed_line
--- a/tests/data/commented_out_magic.ipynb
+++ b/tests/data/commented_out_magic.ipynb
@@ -1,0 +1,55 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "[1, 2,\n",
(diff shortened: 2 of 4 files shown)


nbqa/save_source.py
```diff
@@ -18,7 +18,7 @@
 if TYPE_CHECKING:
     from pathlib import Path
 
-CODE_SEPARATOR = "# %%"
+CODE_SEPARATOR = "# %%\n"
```
MarcoGorelli, line 21: this doesn't make a difference in this file, but `CODE_SEPARATOR` is imported in `nbqa/replace_source.py` and is used to split the cells. We need to include the newline in the splitter else a commented-out cell magic will be split into two different cells

MarcoGorelli:
virtually certain this is fine - tests pass, tried it out on pymc3 notebooks, and it's fine, merging so this bug is fixed

nbqa/save_source.py
s-weigand, line 21: It might be better to make this more 'nbqa unique', e.g. `# %%NBQA-CELL-SEP\n`.
I can imagine a case where `# %%\n` would end up in a real word notebook, e.g. the user doesn't remember the name of a cell magic on the spot and left the start of the cell magic as a comment, just as a reminder to use implement it later.

MarcoGorelli: that's a good point actually, thank you so much!

PR #525, opened 2020-12-28
for flake8, save token as string

### LINKED ISSUE #522: wemake-python-styleguide finds false positives
```
WPS432 Found magic number: 355033482
```

this is when running it on notebook_for_testing.ipynb

MarcoGorelli:
Not my favourite solution to special-case a command, but I can't think how else to do this

closes #522


girip11 (approved):

MarcoGorelli:
cool, thanks @girip11 for reviewing!

PR #541, opened 2021-02-17
Remove 'bdist_wheel universal=True' since the package is Python 3.6.1+

s-weigand:
This is just a minor cleanup for the [filename of the generated wheel](https://pypi.org/project/nbqa/0.5.7/#files), which includes `py2` and might confuse users.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,9 +59,6 @@ nbqa =
 exclude =
     tests*
 
-[bdist_wheel]
-universal = True
-
 [flake8]
 extend-ignore = E203,E503
 max-line-length = 120


MarcoGorelli:
Thanks @s-weigand - doesn't this signal that the project has no compiled extensions? Is the reason to remove just so that `py2` doesn't show up in the versions for the wheel?

s-weigand:
The [universal bdist_wheels](https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels) indicates that the wheel is for python2 and python3 **AND** has no compiled extension.

And yeah the only difference will be `py2` not being in the filename since `python_requires` already tells pip that it doesn't work with python2. So yeah it is just a minor beauty thing 😝

MarcoGorelli:
always up for beauty things - thanks!

PR #542, opened 2021-02-17
Make mutations atomic

s-weigand:
Depending on how many cells and plots a notebook has it can be quite big, thus writing mutations may take some time.
If writing the mutations gets interrupted for some reason (e.g. `KeyboardInterrupt`) this might corrupt the notebook.
In the best-case scenario, the user just loses uncommitted changes, and in the worst-case, their whole work is gone.

Thus writing the changes to a temp file and overwriting the original by a simple `move` reduces the risk of data corruption.

Note that I used `shutil.move()` instead of `Path.rename()` [because `move` also works if the temp and original file are on different drives](https://stackoverflow.com/a/16845955/3990615).

And ofc Anthony  also has a video on this 😋 
[what is atomicity (intermediate) anthony explains #168](https://www.youtube.com/watch?v=tmlmCG5egI0 "what is atomicity (intermediate) anthony explains #168")

### DIFF
--- a/nbqa/replace_source.py
+++ b/nbqa/replace_source.py
@@ -7,6 +7,7 @@ The converted file will have had the third-party tool run against it by now.
 import json
 import sys
 from difflib import unified_diff
+from shutil import move
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
Hey @s-weigand - thanks for doing this, really appreciate it.

Just wondering - any idea why Anthony doesn't do this himself? E.g. 

https://github.com/asottile/pyupgrade/blob/49b8cba72e2222451be96402428d771f0b6fe1ea/pyupgrade/_main.py#L844-L846

(though I may have misunderstood something)

s-weigand:
Hmmmmm 🤔 

Maybe this isn't an issue with python files so much, since they mostly are in the range of a few KB.
But notebooks can be a few MB (e.g. [this 3MB pymc notebook](https://github.com/pymc-devs/pymc-examples/blob/main/examples/getting_started.ipynb)), so there is a higher possible risk.

Also, that part of the code is 4y old. 
I don't know about you, but I always think "who wrote this?" when I look at my old code 😛 
So maybe this is a "no need to fix what isn't broken" artifact. 😉

MarcoGorelli:
That's a good point, this is more of an issue for heavy ipynb files - merging then, thanks, much appreciated!

PR #558, opened 2021-04-02
Refactor ignoring cells

MarcoGorelli:
If we just ignore empty cells this lets us simplify a few things


MarcoGorelli:
merging so I can rebase `opt-in` onto this, will keep that one open longer

Issue #500, opened 2020-12-05
Black removing first character of the first code cell in nb


ncoop57:
Hi, I am using `nbqa==0.5.3` and `black==20.8b1` and have the following notebook:

![image](https://user-images.githubusercontent.com/7613470/101266948-a95fac00-3721-11eb-97d7-176626a85104.png)

Here is a link to the notebook if that helps: https://drive.google.com/file/d/1EyZlfuoAbQycPYY92ysJ03-5lBuE0VlE/view?usp=sharing

As you can see the first cell has this special comment I need for another library I use. When I run `nbqa black 00_core.ipynb --nbqa-mutate`, I get the following output:

![image](https://user-images.githubusercontent.com/7613470/101266974-00658100-3722-11eb-8ed4-437bc4d6e8f8.png)

For some reason, it deletes the comment symbol. Any idea why this is happening or if this is expected? Any help would be greatly appreciated, thanks!

BTW, really like the project!

ncoop57:
Also, this doesn't for some reason happen if I places a cell with a number in it before the cell with the comment symbol:

![image](https://user-images.githubusercontent.com/7613470/101267026-8f729900-3722-11eb-879c-718eca907d2b.png)

Which is what I've been having to do, but its ugly :(

MarcoGorelli:
Hey @ncoop57 

Thanks for the report! This is not expected and is something I wasn't aware of - I'll look into it and'll get back to you when I have a fix!

> BTW, really like the project!

Thanks, and thanks for reporting a bug (which helps make it better!)

MarcoGorelli:
:thinking: I can't reproduce this:
```console
$ nbqa black 00_core.ipynb --nbqa-mutate
All done! ✨ 🍰 ✨
1 file left unchanged.
```

Is this the exact notebook you ran? Do you have a pyproject.toml file in your repo with any settings set?

girip11:
> I can't reproduce this
 Same here.

ncoop57:
Whoops, sorry about that. The file I shared didn't have improper formatting and so black didn't format it. Here is a version that will get formatted by black: https://drive.google.com/file/d/1EyZlfuoAbQycPYY92ysJ03-5lBuE0VlE/view?usp=sharing

> Is this the exact notebook you ran? Do you have a pyproject.toml file in your repo with any settings set?

[...]

PR #559, opened 2021-04-02
opt-in to cell magics

### LINKED ISSUE #557: Make cell-magics opt-in rather than opt-out
Currently, nbqa tries to process all cell magics _except_ those in 

https://github.com/nbQA-dev/nbQA/blob/a99e0e2413c7af340ba994f86752bb29b8b36a82/nbqa/save_source.py#L31-L51

I think we should do the opposite: _never_ process cell magics, except for some hard-coded common ones like `matplotlib inline` and `timeit`, + add a command-line argument so users can pass extra ones they might want to process

This means deprecating `nbqa_ignore_cells` and instead adding `nbqa_process_cells`.

IMO it's OK to make a 0.6.0 release with this breaking change in - this is a relatively new project, and it's not like the user-base is large enough that we need to go through a deprecation cycle

cc @nbQA-dev/nbqa-core in case anyone has any comments - else I'll try to work on during the weekend

MarcoGorelli:
closes #557


girip11:
@MarcoGorelli : I am will review this PR tomorrow. Could you keep this open till then?

nbqa/save_source.py
```diff
@@ -305,7 +285,7 @@ def _should_ignore_code_cell(
     ----------
     source
         Source from the notebook cell
-    ignore_cells
+    process_cells
         Extra cells which nbqa should ignore.
```
girip11, line 289: `Extra cells which nbqa should ignore.` should be updated to `Extra cell magics that nbqa should process.`

nbqa/save_source.py
```diff
@@ -315,14 +295,18 @@ def _should_ignore_code_cell(
...
     if not source:
         return True
-    ignore = MAGIC + [i.strip() for i in ignore_cells]
-    return any(source[0].lstrip().startswith(i) for i in ignore)
+    process = MAGIC + [i.strip() for i in process_cells]
+    magic_type = MagicHandler.get_ipython_magic_type(source[0])
+    if magic_type != IPythonMagicType.CELL:
+        return False
+    first_line = source[0].lstrip()
+    return all(first_line.split()[0] != f"%%{i}" for i in process)
```
girip11, line 303: `all(first_line.split()[0] != f"%%{i}" for i in process)` could be simplified as `first_line.split()[0].lstrip("%%") not in process`

nbqa/save_source.py
```diff
@@ -334,7 +318,7 @@ def main(
         Jupyter Notebook third-party tool is being run against.
     temp_python_file
         Temporary Python file to save converted notebook in.
-    ignore_cells
+    process_cells
         Extra cells which nbqa should ignore.
```
girip11, line 322: Update the description of `process_cells` parameter

tests/tools/test_mypy_works.py
```diff
@@ -45,7 +45,7 @@ def test_mypy_works(capsys: "CaptureFixture") -> None:
         {path_0}:cell_2:19: error: Argument 1 to "hello" has incompatible type "int"; expected "str"
         {path_3}:cell_8:3: error: Name 'flake8_version' is not defined
         {path_3}:cell_8:4: error: Name 'flake8_version' is not defined
-        Found 5 errors in 4 files (checked 20 source files)
+        Found 5 errors in 4 files (checked 22 source files)
```
girip11, line 48: Instead of updating this count every time, shall we do something like `rglob` on `*.ipynb` on the `tests/data` directory?

girip11 (changes requested):

nbqa/save_source.py
MarcoGorelli: Thanks! Though note that `.lstrip("%%")` is the same as `.lstrip("%")`. I've changed this slightly as `lstrip` is really unintuitive, and using it resulted in bug #500

tests/tools/test_mypy_works.py
MarcoGorelli: yeah, definitely :) will keep that for a separate PR though as this one is big enough

girip11: Sure, I just pointed out.

PR #561, opened 2021-04-02
Notebook shadown module causing mypy errors

### LINKED ISSUE #556: notebook shadow module causing mypy errors
When a package contains both notebook and a module with the same exact name, it seems like nbqa-mypy gets confused and the notebook shadows the module causing mypy errors.

Example structure:

```
pipelines/foo
├── __init__.py
├── bar.py
├── bar.ipynb
```

My config:

```yaml
  - repo: https://github.com/nbQA-dev/nbQA
    rev: 0.5.9
    hooks:
      - id: nbqa-mypy
        args: ["--nbqa-ignore-cells=%%R,%%bash", "--strict", "--show-error-codes", "--pretty", "--show-error-context"]
```

MarcoGorelli:
closes #556


nbqa/__main__.py
```diff
@@ -38,7 +39,7 @@
     r"_build|buck-out|build|dist|venv"
     r")/"
 )
-
+TOKEN = secrets.token_hex(3)
```
girip11, line 42: Minor: Instead of `secrets.token_hex` should we go with `hash(notebook_file_path)` so that its reproducible in  case of bug reports?

tests/tools/test_mypy_works.py
```diff
@@ -74,3 +74,17 @@ def test_mypy_with_local_import(capsys: "CaptureFixture") -> None:
     out, _ = capsys.readouterr()
     expected = "Success: no issues found in 1 source file\n"
     assert out == expected
+
+
+def test_notebook_doesnt_shadow_python_module(capsys: "CaptureFixture") -> None:
+    """Check that notebook with same name as a Python file doesn't overshadow it."""
+    cwd = os.getcwd()
+    try:
```
girip11, line 82: Instead of `try..finally` how about writing a context manager

```python
from contextlib import contextmanager
@contextmanager
def change_dir(cwd: str):
    os.chdir(os.path.join("tests", "data"))
    yield
    os.chdir(cwd)

with change_dir(cwd):
    with pytest.raises(SystemExit):
                main(["mypy", "t.ipynb"])
```

I find using context manager more relevant compared to `try..finally` semantically.

girip11 (changes requested):

tests/tools/test_mypy_works.py
MarcoGorelli: Does this work? Because if the `main` command throws an error, we won't reach `os.chdir(cwd)`, while `finally` is always executed

nbqa/__main__.py
MarcoGorelli: good idea, thanks!

tests/tools/test_mypy_works.py
s-weigand: How about making it a fixture with `try..finally` for maximum reusability?

MarcoGorelli: will do if this needs doing elsewhere too, else IMO it's simple enough like this

girip11 (approved):

PR #609, opened 2021-06-08
More robust checking for cell magics

MarcoGorelli:
still not keen on `_should_ignore_cell`, but...getting there


MarcoGorelli:
@scottclowe I noticed your issue in pre-commit about using `black` on a notebook and processing magics - just in case it's useful to you, nbQA will do that in a robust way (using `IPython.core.inputtransformer2.TransformerManager` and `ast`), and has pre-commit hooks defined

scottclowe:
Oh, great! Thanks for letting me know, I hadn't heard of nbQA before. This seems really useful.

PR #618, opened 2021-06-25
BUG allow local scripts

MarcoGorelli:
couldn't figure out how to write an automated test for this, should come back to it later


MarcoGorelli:
> couldn't figure out how to write an automated test for this, should come back to it later

fortunately, coverage doesn't allow me to get away with lazy-ass programming

Issue #626, opened 2021-07-20
Different behaviour between `nbqa black` and `nbqa black --check`


tcbegley:
Hello,

Thanks for making this awesome tool.

I've been having some problems related to `nbqa black --check` raising problems but `nbqa black` not doing anything to change it. I've managed to recreate this with the following minimal example. It appears to be related to how the configuration is parsed.

```
.
├── notebook.ipynb
└── pyproject.toml
```

My project has something like the following configuration. We set the line length to be 79 for source code but are happy with longer lines inside notebooks.

```toml
# pyproject.toml
[tool.black]
line-length = 79

[tool.nbqa.addopts]
# use longer lines in notebooks
black = [
    "--line-length=88",
]
```

`notebook.ipynb` has a single cell with the following contents

```python
result = function_name(xxxxxxxxxxxxxxxxxxxxxxxxxx, xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx)
```

The exact contents are not so important, the important thing is that the line length is more than 79 but less than 88. Running

```sh
nbqa black notebook.ipynb
```

results in no changes, but running

```sh
nbqa black --check notebook.ipynb
```

indicates that the notebook would be reformatted.

![image](https://user-images.githubusercontent.com/15220906/126333704-1345f9de-90bd-4e0c-a8b6-6b6cc1fb5106.png)

To avoid any ambiguity, here's the notebook JSON

```json
{
 "cells": [
  {
   "cell_type": "code",
   "execution_count": null,
   "id": "77402ebb-99fc-4232-9f9a-d46bb0d3aa3d",
   "metadata": {},
   "outputs": [],
   "source": [
    "result = function_name(xxxxxxxxxxxxxxxxxxxxxxxxxx, xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx)"
   ]
  }
 ],
 "metadata": {
  "kernelspec": {
   "display_name": "nbqa",
   "language": "python",
   "name": "nbqa"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 3
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython3",
   "version": "3.9.5"
  }
 },
 "nbformat": 4,
 "nbformat_minor": 5
}
```

tcbegley:
[...]

PR #627, opened 2021-07-20
Extend addopts rather than overwrite

tcbegley:
This PR allows addopts to be extended with command line arguments rather than overridden, fixing #626 

I added a simple test, actually it modifies an existing test that seemed to be checking for the overwrite behaviour. Let me know if you'd rather have a new test and whether I chose an appropriate location for the test.

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -325,8 +325,7 @@ def _get_configs(cli_args: CLIArgs, project_root: Path) -> Configs:
         if getattr(cli_args, section) is not None:
             if section == "addopts":
                 # addopts are added to / overridden rather than replaced outright
-                # Sequence[str] has no attribute extend. Should this be typed as List[str]?
-                config["addopts"].extend(getattr(cli_args, section))  # type: ignore
+                config["addopts"] = (*config["addopts"], *getattr(cli_args, section))
             else:
                 # TypedDict key must be a string literal
                 config[section] = getattr(cli_args, section)  # type: ignore
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -323,8 +323,13 @@ def _get_configs(cli_args: CLIArgs, project_root: Path) -> Configs:
     # If a section was passed via CLI, use that.
     for section in config:
         if getattr(cli_args, section) is not None:
-            # TypedDict key must be a string literal
-            config[section] = getattr(cli_args, section)  # type: ignore
+            if section == "addopts":
+                # addopts are added to / overridden rather than replaced outright
+                # Sequence[str] has no attribute extend. Should this be typed as List[str]?
+                config["addopts"].extend(getattr(cli_args, section))  # type: ignore
+            else:
+                # TypedDict key must be a string literal
+                config[section] = getattr(cli_args, section)  # type: ignore
 
     # add default options
     if cli_args.command == "isort":
--- a/tests/test_pyproject_toml.py
+++ b/tests/test_pyproject_toml.py
@@ -72,12 +72,13 @@ def test_cli_extends_pyprojecttoml(capsys: "CaptureFixture") -> None:
 
     # if arguments are specified on command line, they will take precedence
     # over those specified in the pyproject.toml
+    notebook_path = os.path.join("tests", "data", "notebook_for_testing.ipynb")
     expected_out = dedent(
-        """\
-        tests/data/notebook_for_testing.ipynb:cell_1:1:1: F401 'os' imported but unused
-        tests/data/notebook_for_testing.ipynb:cell_1:3:1: F401 'glob' imported but unused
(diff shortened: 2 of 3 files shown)


nbqa/__main__.py
```diff
@@ -323,8 +323,13 @@ def _get_configs(cli_args: CLIArgs, project_root: Path) -> Configs:
     # If a section was passed via CLI, use that.
     for section in config:
         if getattr(cli_args, section) is not None:
-            # TypedDict key must be a string literal
-            config[section] = getattr(cli_args, section)  # type: ignore
+            if section == "addopts":
+                # addopts are added to / overridden rather than replaced outright
+                # Sequence[str] has no attribute extend. Should this be typed as List[str]?
+                config["addopts"].extend(getattr(cli_args, section))  # type: ignore
```
tcbegley, line 329: As noted in the comment, because `config["addopts"]` is typed as a `Sequence[str]`, Mypy doesn't like the usage of `extend`. Should it be typed as `List[str]` instead? I haven't checked yet if it could ever be anything other than a list.

tests/test_pyproject_toml.py
```diff
@@ -40,7 +40,7 @@ def test_pyproject_toml_works(capsys: "CaptureFixture") -> None:
     assert out == expected_out
 
 
-def test_cli_overwrites_pyprojecttoml(capsys: "CaptureFixture") -> None:
+def test_cli_extends_pyprojecttoml(capsys: "CaptureFixture") -> None:
```
tcbegley, line 43: Should this existing test be modified like this or should I make a new test and leave the old one (it still passes, but the semantics of the name might be confusing)?

Also is this an appropriate location for the test?

nbqa/__main__.py
tcbegley: Having taken a quick look, it seems the default is an empty list, and will only be something else if overwritten by the parsed `pyproject.toml`.

I think the only way there could be a problem is if the `pyproject.toml` itself is misspecified, e.g. if someone did something like

```toml
[tool.nbqa.addopts]
black = "invalid"
```

In that case you probably have bigger problems, so perhaps the more explicit typing would be ok? Please tell me if there's something I missed though.

MarcoGorelli: Would something like
```
config["addopts"] = (*config["addopts"], *getattr(cli_args, section))
```
so that we don't lose immutability in the rest of the codebase?

tests/test_pyproject_toml.py
MarcoGorelli: Yeah, if the old test's name/behaviour no longer make sense, I think it's OK to overwrite

MarcoGorelli:
Thanks @tcbegley for your awesome PR! I'll take a closer look / merge / make a new release after work

MarcoGorelli (changes requested):

nbqa/__main__.py
tcbegley: Sure, have changed as you suggested. 

To clarify though the value of `config["addopts"]` was previously a (mutable) list, are you referring to not using in-place operations such as `.extend` rather than avoiding mutable data structures?

tests/test_pyproject_toml.py
```diff
@@ -64,11 +64,20 @@ def test_cli_overwrites_pyprojecttoml(capsys: "CaptureFixture") -> None:
...
-    out, _ = capsys.readouterr()
-    expected_out = ""
+    # if arguments are specified on command line, they will take precedence
+    # over those specified in the pyproject.toml
+    expected_out = dedent(
+        """\
+        tests/data/notebook_for_testing.ipynb:cell_1:1:1: F401 'os' imported but unused
+        tests/data/notebook_for_testing.ipynb:cell_1:3:1: F401 'glob' imported but unused
+        tests/data/notebook_for_testing.ipynb:cell_1:5:1: F401 'nbqa' imported but unused
+        tests/data/notebook_for_testing.ipynb:cell_4:1:1: F401 'random.randint' imported but unused
```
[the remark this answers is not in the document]
tcbegley: Thanks! Didn't occur to me that would be a problem when the tests passed locally, obviously living in a bubble 😅

tests/test_pyproject_toml.py
tcbegley: I was thinking again about the old test that checked CLI arguments overrode `pyproject.toml` configuration. That is probably still behaviour that should be checked for arguments to nbQA rather than the chosen tool.

E.g. perhaps there should be a test that applies e.g. `--nbqa-skip-cell-tags` at command line and also in `pyproject.toml` and verifies that CLI takes precedence?

Let me know if you want to include something like that as part of this PR or whether it can wait for follow-up work.

MarcoGorelli (approved):

MarcoGorelli:
@all-contributors please add @tcbegley for idea, code, doc

Issue #343, opened 2020-10-17
Release nbqa via conda-forge


s-weigand:
If you want to add `nbqa` on `conda-forge` I can make a PR for that.

MarcoGorelli:
That would be awesome, thank you!

s-weigand:
I opened a [PR at conda-forge](https://github.com/conda-forge/staged-recipes/pull/12923)

s-weigand:
Since the PR got merged and we have 3 releases on conda-forge, I will close this issue.
If you want to add a badge, here is the md code for that ^^
[![Conda Version](https://img.shields.io/conda/vn/conda-forge/nbqa.svg)](https://anaconda.org/conda-forge/nbqa)
```markdown
[![Conda Version](https://img.shields.io/conda/vn/conda-forge/nbqa.svg)](https://anaconda.org/conda-forge/nbqa)
```

PR #684, opened 2021-12-02
Add conda installation information

asdfCYBER:
Update documentation to include conda installation information, because nbQA is available on the conda-forge channel since #343.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,12 @@
 		alt="docs"
 		src="https://readthedocs.org/projects/nbqa/badge/?version=latest"></a>
 </p>
+
 <p align="center">
 	<a href="https://pepy.tech/project/nbqa"><img
 		alt="downloads"
 		src="https://pepy.tech/badge/nbqa"></a>
 </p>
-
 
 <p align="center">
     <a href="#readme">
--- a/README.md
+++ b/README.md
@@ -32,6 +32,12 @@
 		alt="docs"
 		src="https://readthedocs.org/projects/nbqa/badge/?version=latest"></a>
 </p>
+<p align="center">
+	<a href="https://pepy.tech/project/nbqa"><img
+		alt="downloads"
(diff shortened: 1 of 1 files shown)


s-weigand (approved):
LGTM 👍

MarcoGorelli:
Thanks - TBH I don't really feel comfortable maintaining the conda package, so I'm OK with this being on conda but I'm tempted to not advertise it

MarcoGorelli:
Just out of interest, what's the advantage of downloading this tool from conda-forge (as opposed to from pip)?

s-weigand:
The thing with conda and pip is that mixing them can mess things up (mostly for packages with c-extensions).
Also when you install it with pip `conda remove` doesn't work so users have to look up with which package manager they installed what.
And one more thing with `coda` that the dependency compatibility resolution is a lot stricter than the one pip uses.

asdfCYBER:
> Thanks - TBH I don't really feel comfortable maintaining the conda package, so I'm OK with this being on conda but I'm tempted to not advertise it

I'm not familiar with the differences between pip and conda distributions, but isn't the version on conda basically a copy of the latest pip version? (basing this off https://github.com/conda-forge/staged-recipes/pull/12923#issuecomment-712148421). AFAIK that'd mean that as long as the version on PyPI works the conda-forge package does not require additional work.

As for the advantage of downloading from conda-forge instead of PyPI, pretty much what s-weigand said. Using pip in conda environments bypasses conda so that it doesn't have a complete list of which packages are installed anymore, which can in rare cases break the entire environment. See also https://www.anaconda.com/blog/using-pip-in-a-conda-environment.

MarcoGorelli (approved):
Sure, let's do this (just checked, and to be fair, `black` is available on `conda-forge`, so if that's anything to go by, so should nbQA)

MarcoGorelli:
@all-contributors please add @asdfCYBER for docs

PR #693, opened 2022-01-11
unpin

MarcoGorelli:


setup.cfg
```diff
@@ -36,7 +36,7 @@ py_modules = nbqa
 install_requires =
     ipython>=7.8.0
     tokenize-rt>=3.2.0
-    tomli<2.0.0
+    tomli
```
MarcoGorelli, line 39: I should never have accepted this upper-bound, but whatever, too late now, all I can do is remove it

s-weigand: Better upper bounds and testing that it works than stuff that breaks all of the sudden all over the place 😝

MarcoGorelli: dunno, the risk is that then `pip` doesn't let people install things because of unnecessary constraints their packages have specified

PR #717, opened 2022-04-04
Update Black version to 22.3.0

### LINKED ISSUE #716: black being pinned to an older version causes a crash
There was a breaking change recently in the `click` library which is a dependency in `black`. `black` has updated the dependency version and released a new version, however because `black` is pinned to an older version, this fix isn't getting propagated. For the moment, we can workaround this by pinning `click` to an older version, but I think `nbqa-black` itself can fix this by moving to the newest black version.

I checked the commit that introduced the version pin, but I couldn't figure out the reason for doing so.

Refer to the [upstream issue](https://github.com/psf/black/issues/2964) for more details.

chamilad:
This change updates the precommit config to have the latest Black
version (at the time of this commit). The main reason for doing so is to
get the latest fixes for a bug caused in a dependency library, click.
More details about this can be found at
https://github.com/psf/black/issues/2964

This resolves #716

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     hooks:
       - id: validate_manifest
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
       - id: black
   - repo: https://github.com/pre-commit/mirrors-prettier


chamilad:
pylint.org was down when this ran, but they seems to have added a redirect to [pylint.pyqca.](https://pylint.pycqa.org/). Should be good if the workflow is run again.

MarcoGorelli:
Thanks @chamilad 

Closing as the solution is to pin `black` in `additional_dependencies`

s-weigand:
> Thanks @chamilad
> 
> Closing as the solution is to pin `black` in `additional_dependencies`

While this does not fix #716 , it fixes that pre-commit is broken when you don't have a cached env for `black` with `click<8.1.0`.
Additionaly, the [pinned `black` version in reqirements_dev.txt](https://github.com/nbQA-dev/nbQA/blob/b77ee1424b4ef11878c99ebd8a701ed442fa0aa8/requirements-dev.txt#L3) should also be raised to `22.3.0` since it [currently breaks the CI](https://github.com/nbQA-dev/nbQA/runs/5812082005?check_suite_focus=true) (guess that also needs some expected test values to change).
#715 adds this change and more, but due to the broken CI the tests fail.

The recent pallets x.1.0 releases really were anti semantic versioning ones, first `jinja2` 3.1.0 breaks `nbconvert` (RTD now pins `jinja2<3.1.0` in their docker images) and after that `click` 8.1.0 breaks `black` and `typer`. I also had a lot of updates to do all over the place 😅

PR #724, opened 2022-06-08
ENH: add nbqa-shell

### LINKED ISSUE #695: flakeheaven
I would love to run note plain flake8 but flake8 as wrapped by [flakeheaven](https://github.com/flakeheaven/flakeheaven). Alas, that appears to be difficult [given how nbqa is currently architected](https://github.com/flakeheaven/flakeheaven/issues/17). Can you think of a fix for this that the authors of flakeheaven haven't yet thought of. Alternatively, would you be willing to go down any of the routes proposed in the linked issue?

pwoolvett:
Enable a new flag `--nbqa-shell`. When
set, instructs nbqa to run the command argument directly,
instead of importing it as a module via python.

* updated docs in examples to include reference usage.
* updated cli usage message to include new flag.
* add new exception used when the requested script is not found.
* update interal api when shell is required:
    * __main__._run_command
    * __main__._check_command_is_installed
    * cmdline.CLIArgs
    * config.config.Configs
    * config.config.get_default_config
* added tests for cli flag:
    * Fallback to default usage when flag not specified.
    * Use `sp.run` directly instead of `python -m` usage when flag is set.
    * raise exception when flag is passed and command is not found.

fixes #695

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -7,6 +7,7 @@ import sys
 import tempfile
 from importlib import import_module
 from pathlib import Path
+from shutil import which
 from textwrap import dedent
 from typing import (
     Any,
@@ -81,6 +82,15 @@ class UnsupportedPackageVersionError(Exception):
             f"{BOLD}nbqa only works with {command} >= {min_version}, "
             f"while you have {current_version} installed.{RESET}"
         )
+        super().__init__(self.msg)
+
+
+class CommandNotFoundError(Exception):
+    """Raise if requested command cannot be found in $PATH."""
+
+    def __init__(self, command: str) -> None:
+        """Initialise with command."""
+        self.msg = f"{BOLD}nbqa was unable to find {command}.{RESET}"
         super().__init__(self.msg)
 
 
@@ -226,6 +236,8 @@ def _run_command(
     command: str,
     cmd_args: Sequence[str],
     args: Sequence[str],
+    *,
+    shell: bool,
 ) -> Tuple[Output, int, bool]:
     """
     Run third-party tool against given file or directory.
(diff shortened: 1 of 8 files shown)


nbqa/__main__.py
```diff
@@ -224,6 +234,7 @@ def _get_mtimes(arg: str) -> Set[float]:
 
 def _run_command(
     command: str,
+    shell: bool,
     cmd_args: Sequence[str],
     args: Sequence[str],
 ) -> Tuple[Output, int, bool]:
```
MarcoGorelli, line 240: [boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap), let's rewrite as:
```suggestion
def _run_command(
    command: str,
    cmd_args: Sequence[str],
    args: Sequence[str],
    *,
    shell: bool,
) -> Tuple[Output, int, bool]:
```

MarcoGorelli (changes requested):
Nice, thanks for your PR!

I need to fixup some unrelated CI failures, then I'll come back to this

MarcoGorelli (changes requested):
Thanks for updating

I've just merged a PR fixing some of the upstream issues, if you rebase/merge then this should become green

MarcoGorelli (approved):
Awesome, thanks @pwoolvett ! Should be good to go

I'm just about to go on holiday, so I'll give it a final check / merge/ make release when I'm back

MarcoGorelli:
Hi @pwoolvett  - reckon this option needs to be configurable from `pyproject.toml`?

MarcoGorelli:
Fixed this up a bit so it's configurable from pyproject.toml - thanks for your PR!

PR #732, opened 2022-07-11
Document unsupported `--per-file-ignores` `flake8` flag

### LINKED ISSUE #730: `flake8` `per-file-ignores` not read from `setup.cfg`?
```cfg
[flake8]
# general ignore works
ignore = E402
per-file-ignores =
  # per-file ignore doesn't work
  examples/*.ipynb: E402
```

If I remove `ignore = E402`, I still get

```sh
$ nbqa flake8 examples/*.ipynb
                  
examples/foo.ipynb:cell_9:2:1: E402 module level import not at top of file
```

if I leave it in, the error disappears so `setup.cfg` is read but `per-file-ignores` apparently isn't?

janosh:
Closes #730.


MarcoGorelli:
Thanks @janosh !

PR #804, opened 2023-03-19
autoupdate

MarcoGorelli:


nbqa/__main__.py
```diff
@@ -43,6 +32,18 @@
 from nbqa.save_code_source import CODE_SEPARATOR
 from nbqa.text import BOLD, RESET
 
+
+def parse_version(version: str) -> tuple[int, ...]:
+    """
+    Parse version; split into a tuple of ints for comparison.
+
+    Borrowed from polars, I can't tell what Python expects us to use.
+    pkg_resources is deprecated apparently.
```
gaborbernat, line 41: https://pypi.org/project/packaging/ is what you should be using instead...

MarcoGorelli: thanks!

PR #807, opened 2023-04-01
pass autopep8 first

### LINKED ISSUE #806: False positive for `ruff` `I001` (isort) when there is a function/class definition after the imports cell
This bug is similar to #796 (thank you very much for fixing that!).
It arises when running `nbqa ruff`

- on a notebook with any import in the first cell
- where the isort checks are enabled e.g. with --select=I
- The cell following the cell containing imports starts with a class or function definition

As in #796, there is a false positive error message of 

```
I001 [*] Import block is un-sorted or un-formatted
```

and `--fix` does not change the notebook, even though it reports it as fixed.

My minimal notebook has the following cells:

```python
import os
```

```python
class A:
    ...
```

If I add the `--diff` option, it wants to insert a newline

```diff
@@ -1,6 +1,7 @@
 # %%NBQA-CELL-SEPfc4e93
 import os

+
 # %%NBQA-CELL-SEPfc4e93
 class A:
     ...

Would fix 1 error.
```

MarcoGorelli:
closes #806


nbqa/__main__.py
```diff
@@ -308,6 +306,14 @@ def _run_command(
     else:
         python_module = COMMAND_TO_PYTHON_MODULE.get(main_command, main_command)
         cmd = [sys.executable, "-m", python_module, *sub_commands]
+    # fixup line breaks
+    subprocess.run(
+        [sys.executable, "-m", "autopep8", "--select=E3", "--in-place", *args],
+        capture_output=True,
+        text=False,
+        env=my_env,
+    )
```
MarcoGorelli, line 315: probably need to record what changes it makes, and preserve them

nbqa/replace_source.py
```diff
@@ -91,7 +91,12 @@ def _reinstate_magics(
         New source that can be saved into Jupyter Notebook.
     """
     for magic_substitution in temporary_lines:
-        source = source.replace(magic_substitution.replacement, magic_substitution.src)
+        source = re.sub(
+            f"{re.escape(magic_substitution.replacement)}\n*",
```
MarcoGorelli, line 95: instead of subbing any number of newlines, let's try to preserve the original number

tests/test_ipython_magics.py
```diff
@@ -45,12 +45,14 @@ def _validate_magics_with_black(before: Sequence[str], after: Sequence[str]) ->
     diff = difflib.unified_diff(before, after)
     result = "".join(i for i in diff if any([i.startswith("+ "), i.startswith("- ")]))
     expected = (
+        '-    "\\n",\n'
```
MarcoGorelli, line 48: this happens from removing the newline after the magic. need to decide whether to standardise, or whether this is ok. maybe it's OK

tests/test_ipython_magics.py
```diff
@@ -45,12 +45,14 @@ def _validate_magics_with_black(before: Sequence[str], after: Sequence[str]) ->
...
         '-    "def compute(operand1,operand2, bin_op):\\n",\n'
         '+    "def compute(operand1, operand2, bin_op):\\n",\n'
         '-    "compute(5,1, operator.add)"\n'
         '+    "compute(5, 1, operator.add)"\n'
         '-    "    ?str.splitlines"\n'
         '+    "str.splitlines?"\n'
+        '-    "\\n",\n'
         '-    "   %time randint(5,10)"\n'
```
MarcoGorelli, line 56: this one comes from
```
     "flake8_version = !pip list 2>&1 | grep flake8\n",
-    "\n",
     "if flake8_version:\n",
     "    my_var = flake8_version"
```
which...might not be a good idea

prob. best to figure out how to preserve newlines

can probably count new lines before applying autopep8, count them after, and then use that in the substitution?

should be good. might need to expand notebookinfo even further

tests/tools/test_ruff_works.py
```diff
@@ -59,10 +59,10 @@ def test_ruff_works(
...
         f"{expected_path_0}:cell_1:3:8: F401 [*] `glob` imported but unused\n"
         f"{expected_path_0}:cell_1:5:8: F401 [*] `nbqa` imported but unused\n"
-        f"{expected_path_0}:cell_4:1:1: E402 Module level import not at top of file\n"
-        f"{expected_path_0}:cell_4:1:20: F401 [*] `random.randint` imported but unused\n"
-        f"{expected_path_0}:cell_5:1:1: E402 Module level import not at top of file\n"
-        f"{expected_path_0}:cell_5:2:1: E402 Module level import not at top of file\n"
+        f"{expected_path_0}:cell_4:3:1: E402 Module level import not at top of file\n"
+        f"{expected_path_0}:cell_4:3:20: F401 [*] `random.randint` imported but unused\n"
+        f"{expected_path_0}:cell_5:3:1: E402 Module level import not at top of file\n"
+        f"{expected_path_0}:cell_5:4:1: E402 Module level import not at top of file\n"
```
MarcoGorelli, line 65: this isn't quite right...somehow, need to sort out the line number in the sub too

nbqa/__main__.py
```diff
@@ -516,23 +562,66 @@ def _save_code_sources(
...
+            with open(file_name) as fd:
+                for line in fd:
+                    if line == CODE_SEPARATOR:
+                        parsed_cells.append(current_cell)
+                        current_cell = []
+                    current_cell.append(line)
+                parsed_cells.append("".join(current_cell))
+            if parsed_cells:
+                # skip the initial empty one
+                parsed_cells = parsed_cells[1:]
```
MarcoGorelli, line 607: way too complicated, just return `result` from first pass?

MarcoGorelli:
seems fine, let's ship it and see what happens

PR #814, opened 2023-05-04
Fix documentation to enable isort with 'black' profile

florianbaer:
nbqa-isort throws an error when adding as pre-commit when args are not concatenanated


MarcoGorelli (approved):
thanks @florianbaer

MarcoGorelli:
@all-contributors please add @florianbaer for docs

PR #839, opened 2024-03-01
Update messages in tests/tools/test_pylint_works.py

### LINKED ISSUE #838: tests/tools/test_pylint_works.py fails as pylint 3.1.0 updated error messages
[pylint 3.1.0](https://github.com/pylint-dev/pylint/releases/tag/v3.1.0) updated the message for wrong-import-order:

```
The message will report imports as follows:
For "import X", it will report "(standard/third party/first party/local) import X"
For "import X.Y" and "from X import Y", it will report "(standard/third party/first party/local) import X.Y"
```

this caused the tests/tools/test_pylint_works.py test to fail with the current pylint version. The messages in the test should be updated accordingly.

funkeleinhorn:
Pylint 3.1.0 updated the messages of the wrong-import-order check:
https://github.com/pylint-dev/pylint/releases/tag/v3.1.0

This commit updates the messages in the test accordingly so the test passes again.

Closes #838


MarcoGorelli:
thanks @funkeleinhorn !

PR #870, opened 2024-10-29
add nbqa_ipynb to temporary files

### LINKED ISSUE #789: Temporary files should be created in a temporary directory outside the repository
When nbQA runs it extracts a `foo[randomness].py` file from `foo.ipynb`, and puts this file in the same directory as the notebook. This confuses my IDE, which is set to add any new files to VCS. It would probably be better to put this in a [temporary directory](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp).

minrk:
allows config globs to apply to nbqa-generated files.

Added to suffix, so e.g. all generated .py files match `*nbqa_ipynb.py` e.g. `default_magicgr4dg5bw_nbqa_ipynb.py`

docs and tests updated as best I could find.

closes #789


MarcoGorelli:
@all-contributors please add @minrk for code

PR #881, opened 2025-03-18
Support for running nbqa on quatro format (.qmd) markdown files

hguturu:
Two main changes:
* Added check for '.qmd' when using Jupytex
* Added conversion of leading quartro cell options '#|' that get changed to '# |' by black/ruff back to '#|' 
(`_restore_quarto_cell_options`)

If this is an acceptable way to add support, I can add some test logic to this PR.

### DIFF
--- a/nbqa/__main__.py
+++ b/nbqa/__main__.py
@@ -126,7 +126,7 @@ def _get_notebooks(root_dir: str) -> Iterator[Path]:
         jupytext_installed = True
     if os.path.isfile(root_dir):
         _, ext = os.path.splitext(root_dir)
-        if (jupytext_installed and ext in (".ipynb", ".md")) or (
+        if (jupytext_installed and ext in (".ipynb", ".md", ".qmd")) or (
             not jupytext_installed and ext == ".ipynb"
         ):
             return iter((Path(root_dir),))
@@ -138,7 +138,9 @@ def _get_notebooks(root_dir: str) -> Iterator[Path]:
 
     if jupytext_installed:
         iterable = itertools.chain(
-            Path(root_dir).rglob("*.ipynb"), Path(root_dir).rglob("*.md")
+            Path(root_dir).rglob("*.ipynb"),
+            Path(root_dir).rglob("*.md"),
+            Path(root_dir).rglob("*.qmd"),
         )
     else:  # pragma: nocover
         iterable = itertools.chain(Path(root_dir).rglob("*.ipynb"))
--- a/nbqa/path_utils.py
+++ b/nbqa/path_utils.py
@@ -88,7 +88,7 @@ def read_notebook(notebook: str) -> Tuple[Optional[Dict[str, Any]], Optional[boo
     if ext == ".ipynb":
         trailing_newline = content.endswith("\n")
         return json.loads(content), trailing_newline
-    assert ext == ".md"
+    assert ext in (".md", ".qmd")
     try:
         import jupytext  # pylint: disable=import-outside-toplevel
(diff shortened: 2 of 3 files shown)


hguturu:
@MarcoGorelli PR should be ready for another round of workflow runs by the testsuite. Logic has also been simplified.

malcolmbarrett:
I've been following this, but I noticed while playing with the current implementation that it adds a bunch of metadata to the file's yaml, e.g., in the case of the example in the PR

```yaml
jupyter:
  jupytext:
    text_representation:
      extension: .qmd
      format_name: quarto
      format_version: '1.0'
      jupytext_version: 1.16.7
  kernelspec:
    display_name: Python 3
    language: python
    name: python3
```

Would that be possible to avoid? I don't know how useful it is for the underlying tools, but it certainly takes up a lot of space in an area that often contains important information

hguturu:
That is actually not being done by `nbqa`, but rather `jupytext` and `quarto convert`. 

`nbqa` converts `qmd` to `json` via jupytext at https://github.com/nbQA-dev/nbQA/pull/881/files#diff-d72c6dc865400b582f957459d2f13e4030da8a9f8afdec522def74313d29a25fR109.

And then converts back to `qmd` via jupytext at https://github.com/nbQA-dev/nbQA/pull/881/files#diff-30fd9daaa53315a90c29babeec62d45b8705ebe8f57133a0034a01db15a7860aR229.

The jupytext conversion back to qmd - https://github.com/mwouts/jupytext/blob/main/src/jupytext/jupytext.py#L238.

The `jupytext` metadata is added by https://github.com/mwouts/jupytext/blob/main/src/jupytext/jupytext.py#L185 and the `kernelspec` metadata is added by `quarto convert`.

malcolmbarrett:
Yes, it seems unavoidable without modifying the upstream processes. Thanks for the info!

hguturu:
> @MarcoGorelli PR should be ready for another round of workflow runs by the testsuite. Logic has also been simplified.

Thanks for running the suite. It looks my test case was not robust to changing jupytext version. I have now fixed that.

MarcoGorelli:
Sure thanks - if you want to make a separate pr which just fixes a typo or something then after that ci will run without needing approval

hguturu:
> Sure thanks - if you want to make a separate pr which just fixes a typo or something then after that ci will run without needing approval

Fortunately, I think we are all passing now?

The current failure with docs seems to be related to `doc-links` which is a false positive? Both of the "broken" links seem to be working?

malcolmbarrett:
The code lgtm. 

As for maintenance of the feature, I do think the best outcome is native support for `.qmd` files in `ruff`, not a workaround like this or others I've taken. I don't know if the ruff maintainer will do that or accept a contribution along those lines, but I don't think any approach I've taken has an acceptable impact on the document itself. My question about the metadata earlier is an example. I think a Python formatter/linter should just touch Python code, and it seems inevitable here, in part because of the use of `quarto convert`, which I doubt will ever be able to make that guarantee even if other upstream issues were addressed. 

That's my 2 cents as to whether it's worth maintaining. I have a limited use case (applying ruff to Quarto files), so maybe other users will find this useful for applying other tools to Quarto files.

hguturu:
Yes, I will be available to fix issues if the are reported.

I agree with Malcolm on having the other tools support qmd files natively would be ideal, but this will be a good workaround until that happens. I figure `ruff` will eventually add it since they have this issue open (and the issues for broader jupytext and markdown code block support). Additionally, it is nice for other tools which may never add support.

I also believe the concerns about metadata changes already effect other markdown formats supported? Since myst would also go through jupytext.

PR #906, opened 2026-08-14
Report the first line of a cell as line 1, not line 0

### LINKED ISSUE #821: Invalid line number for reporting diagnostics at first line
## Jupyter notebook content

```python
import math
import random
```

_Cell count: 1_

## Command

```console
$ nbqa --nbqa-shell ruff /path/to/notebook.ipynb --select=ALL
```

***OR***

```console
$ nbqa --nbqa-shell flake8 /path/to/notebook.ipynb
```

### Output

**Note: Output is truncated to only keep the relevant lines**

**Ruff:**
```
...
/Users/dhruv/playground/python/ruff_test.ipynb:cell_1:0:1: D100 Missing docstring in public module
/Users/dhruv/playground/python/ruff_test.ipynb:cell_1:2:8: F401 [*] `random` imported but unused
...
```

**Flake8:**
```
/Users/dhruv/playground/python/ruff_test.ipynb:cell_1:0:1: D100 Missing docstring in public module
/Users/dhruv/playground/python/ruff_test.ipynb:cell_1:2:1: F401 'random' imported but unused
```

Notice that `D100` is reported at line 0 while `F401` is reported at line 2. The former suggests that it's 0-indexed while the latter suggests it's 1-indexed. This is happening because of the `CODE_SEPARATOR` inserted between each code cells:

```python
# %%NBQA-CELL-SEP8e47a4
import math
import random

```

---
* Ruff version: `0.0.269`
* nbqa version: `1.7.0`
* `flake8 --version`:
	```                     
	6.0.0 (flake8-bugbear: 23.5.9, flake8-docstrings: 1.7.0, flake8-pyi: 23.5.0, flake8_implicit_str_concat: 0.4.0, mccabe: 0.7.0, pycodestyle: 2.10.0, pyflakes: 3.0.1) CPython 3.11.3 on Darwin
	```

Eljees:
Closes #821.

A diagnostic about the module as a whole comes back pointing at a line that does not exist:

```console
$ ruff check --isolated --select D100,F401 t.py
t.py:1:1: D100 Missing docstring in public module
t.py:1:8: F401 [*] `math` imported but unused
t.py:2:8: F401 [*] `random` imported but unused

$ nbqa --nbqa-shell ruff t.ipynb --isolated --select D100,F401
t.ipynb:cell_1:0:1: D100 Missing docstring in public module
t.ipynb:cell_1:1:8: F401 [*] `math` imported but unused
t.ipynb:cell_1:2:8: F401 [*] `random` imported but unused
```

`F401` is 1-indexed and `D100` is 0-indexed, in the same output. `flake8` behaves the same way. On the equivalent plain `.py` file both are at line 1, so the 0 does not come from the tool.

### Cause

Every cell in the temporary Python file starts with the `# %%NBQA-CELL-SEP...` comment nbQA inserts itself, and `_get_line_numbers_for_mapping` maps that line to cell line `0`. A module-level message is reported by the tool on line 1 of the file -- that is, on the separator -- so it is mapped back to `cell_1:0`.

This maps the separator line to the first line the user actually wrote.

### Existing expectations

Four `pydocstyle` expectations recorded `cell_1:0 at module level` -- one in `tests/tools/test_pydocstyle.py` and three in `tests/test_jupytext.py`. They froze exactly the behaviour this issue is about, so they are updated to `cell_1:1` in the same commit. Flagging it explicitly rather than burying it in the diff: if you would rather keep the old numbering, this PR is the wrong change and I will close it.

### Tests

In `tests/test_map_python_line_to_nb_lines.py`: a unit test on `_get_line_numbers_for_mapping` and an end-to-end check through `pylint --disable=all --enable=C0114`. Both fail on `main`.

### Test suite

`3 failed, 118 passed, 3 skipped` on this branch. The same three `tests/tools/test_ruff_works.py` failures are present on `main` (newer `ruff` output format) and are unrelated to this change.

### DIFF
--- a/nbqa/save_code_source.py
+++ b/nbqa/save_code_source.py
@@ -234,14 +234,18 @@ def _get_line_numbers_for_mapping(
     lines_in_cell = cell_source.splitlines()
     line_mapping: MutableMapping[int, int] = {}
 
+    # Line 0 of the parsed cell is the CODE_SEPARATOR which nbQA itself
+    # inserted, so a diagnostic reported there belongs to the first line the
+    # user actually wrote. Cell lines are 1-indexed, and cell_N:0 does not
+    # exist - see https://github.com/nbQA-dev/nbQA/issues/821
     if not magic_substitutions:
-        line_mapping.update({i: i for i in range(len(lines_in_cell))})
+        line_mapping.update({i: max(i, 1) for i in range(len(lines_in_cell))})
     else:
         line_number = -1
 
         for line_no, _ in enumerate(lines_in_cell):
             line_number += 1
-            line_mapping[line_no] = line_number
+            line_mapping[line_no] = max(line_number, 1)
 
     return line_mapping
 
--- a/tests/test_jupytext.py
+++ b/tests/test_jupytext.py
@@ -296,11 +296,11 @@ def test_jupytext_on_folder(capsys: "CaptureFixture") -> None:
     )
     out, _ = capsys.readouterr()
     expected = (
-        f'{os.path.join(path, "invalid_syntax.ipynb")}:cell_1:0 at module level:\n'
+        f'{os.path.join(path, "invalid_syntax.ipynb")}:cell_1:1 at module level:\n'
         "        D100: Missing docstring in public module\n"
-        f'{os.path.join(path, "assignment_to_literal.ipynb")}:cell_1:0 at module level:\n'
+        f'{os.path.join(path, "assignment_to_literal.ipynb")}:cell_1:1 at module level:\n'
         "        D100: Missing docstring in public module\n"
-        f'{os.path.join(path, "automagic.ipynb")}:cell_1:0 at module level:\n'
+        f'{os.path.join(path, "automagic.ipynb")}:cell_1:1 at module level:\n'
         "        D100: Missing docstring in public module\n"
     )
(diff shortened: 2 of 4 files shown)


Eljees:
A note on the red marks here, because they say something other than what they look like.

Neither this branch nor #907 has ever been tested. Both runs on the previous head were recorded as failures without starting a single job:

```
run 31778653328  tox    created 2026-08-14 07:06:12Z   updated 2026-09-13 07:06:44Z   conclusion failure   jobs 0
run 31778653242  docs   created 2026-08-14 07:06:12Z   updated 2026-09-13 07:06:44Z   conclusion failure   jobs 0
```

Thirty days and thirty-two seconds apart, which is what happens to a fork run left in `action_required`: the approval window closes and the run is written down as a failure although nothing ran.

I have refreshed both branches onto current `main`. Each now has a fresh `tox` and `docs` run waiting on "Approve and run workflows", so one press gives both pull requests a real result.

One thing to expect when they do run: `tox` is currently red on `main` itself. On #904 it fails three cases in `tests/tools/test_ruff_works.py` with output-text mismatches, which reads as ruff version drift rather than anything in these two branches.

#906 moves a cell's first line from 0 to 1 in reported positions; #907 keeps the tool's exit code when a notebook comes back unchanged. Both are small and independent of each other.

